PR #11, opened 2025-02-20
Update _split.py

aminsoh:
Replaced the previous implementation of _get_list_periodic_pattern_pos() which used a while loop with an optimized implementation using range().


breimanntools:
Thank you for the optimization idea and for digging into the split logic 🙏

Unfortunately the `range()`-based rewrite isn't behavior-preserving. The original interleaves `step1`/`step2` into a single ordered position list and includes positions up to and including `len_seq`, whereas `range(pos, len_seq, stride) + range(pos + step1, len_seq, stride)` returns two concatenated (non-interleaved) ranges and excludes `len_seq`. That changes the periodic-pattern positions, which would alter CPP feature values (guarded by our regression anchor), so it can't be merged as-is.

The current while-loop is correct and isn't a bottleneck here, so we'll keep it. A faithful micro-optimization would need to preserve both the interleaved ordering and the inclusive `len_seq` boundary.

Closing for housekeeping — thanks again for taking the time, contributions are very welcome!

PR #14, opened 2025-04-28
Handle Duplicate Entries in get_df_parts_ Function and Update Documentation

Sadiyahafroz:
This pull request updates the get_df_parts_ function to handle duplicate entries in the COL_ENTRY column. A new parameter handle_duplicates is introduced, allowing the user to specify how duplicates should be handled (either adding serial numbers or leaving them unchanged). If the parameter is not provided, duplicates will be handled by default with serial numbers to ensure unique indexing. Additionally, minor updates were made to the tutorial for clarity.

### DIFF
--- a/aaanalysis/feature_engineering/_backend/cpp/utils_feature.py
+++ b/aaanalysis/feature_engineering/_backend/cpp/utils_feature.py
@@ -186,17 +186,39 @@ def get_list_parts(features=None):
     return list_parts
 
 
-def get_df_parts_(df_seq=None, list_parts=None, jmd_n_len=None, jmd_c_len=None):
-    """Create DataFrame with sequence parts"""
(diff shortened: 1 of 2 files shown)


breimanntools:
Thank you for this contribution and for taking the time to improve `aaanalysis` 🙏

Since this PR was opened the codebase has moved on and it now has merge conflicts. More importantly, the change doesn't fit the package's architecture: `get_df_parts_` is backend code, which by convention trusts the frontend and must not validate/transform raw user input or call `print(...)`, and silently renaming `entry` values by default would change user data without an explicit opt-in.

That said, the underlying point is a real gap — duplicate `entry` values aren't currently validated. We'll address it directly in the frontend (`check_df_seq`) by raising a clear error on duplicate entries, rather than mutating them in the backend. Crediting you for surfacing it.

Closing for housekeeping — please don't let this discourage you from contributing again, and thanks again!

Issue #61, opened 2026-04-11
Multi-class and regression support via random reference sequence generation


breimanntools:
## Problem
CPP is strictly binary: it compares one positive set against one negative/reference
set. This excludes two major categories of real-world problems:

1. **Multi-class tasks** — enzyme family prediction, subcellular localization,
   functional class annotation — where there are more than two groups to compare.
2. **Regression tasks** — thermostability, binding affinity, solubility scores —
   where the target is continuous, not categorical.

A key blocker for both is the absence of a principled reference set when no
natural negative class exists. Random but biologically realistic background
sequences can serve as this reference, unlocking both problem types within the
existing CPP framework.

## Goal
Add a random reference sequence generator that produces background sequences
matching the amino acid composition and length distribution of the input set,
enabling CPP to be applied to multi-class and regression settings without
requiring a curated negative dataset.

### Random reference sequence generator
- [ ] Implement `generate_reference_sequences(sequences, n, method)` supporting:
  - Uniform random shuffling of input sequences (preserves composition)
  - Composition-matched random generation (draws from empirical AA frequencies)
  - Length-matched sampling from UniRef or SwissProt background
- [ ] Allow seeding for reproducibility
- [ ] Validate that generated sequences are statistically similar to input in
      composition and length

### Multi-class support
- [ ] Implement one-vs-rest CPP: each class is compared against the generated
      reference (or a merged background of all other classes)
- [ ] Aggregate per-class feature importance into a unified feature matrix
- [ ] Add multi-class SHAP explanation support (summary across classes)
- [ ] Provide API: `CPP.fit_multiclass(class_dict, reference=None)`

### Regression support
- [ ] Enable CPP to compare high-value vs low-value sequence subsets split at
      a user-defined quantile threshold
- [ ] Alternatively, use the reference generator to create a neutral baseline
      and compute feature deviations correlated with continuous target
- [ ] Add SHAP regression mode with continuous target compatibility
- [ ] Provide API: `CPP.fit_regression(sequences, targets, reference=None)`

## How this improves AAanalysis
- Removes the hard requirement for a curated negative/reference class
- Unlocks multi-class protein function problems within the existing framework
- Opens regression use cases (protein engineering, fitness prediction) without
  changing the core CPP logic
- The reference generator is independently useful as a null model for
  statistical testing

## Acceptance criteria
- Reference generator produces sequences that pass composition similarity checks
- Multi-class CPP produces per-class feature importance consistent with
[...]

Issue #18, opened 2026-04-09
Standardize CPP output schema for downstream usage


breimanntools:
## Problem
CPP outputs are not fully standardized, making integration with:
- ML pipelines
- visualization tools
- ProtXplain

more difficult.

## Goal
Define a consistent and extensible output schema.

## Tasks
- [ ] Define standard dataframe structure (columns, naming)
- [ ] Ensure consistent feature naming
- [ ] Add metadata (e.g., feature type, region)
- [ ] Refactor existing outputs
- [ ] Document schema

## Acceptance criteria
- All CPP outputs follow a unified schema
- Easy integration into ML and visualization pipelines

breimanntools:
Resolved on `master` in **0e561566** (green CI: Unit Tests + Test Coverage + CodeQL).

A `/grill-with-docs` pass reframed the scope: two of the issue's words landed on **reserved glossary terms**, so the issue became *stabilize + document the existing schema* rather than *add metadata columns*:

- **"region"** is reserved (`CONTEXT.md` — the `PART` field is a **part**, not a region; "region" is held for #27). No column added.
- **"feature type"** is ambiguous and already taken by `df_annot.feature_type`. The compositional-vs-positional axis stays *emergent* from `split_kws` and is now documented (it is not a stored column).

**Decided scope:** zero new `df_feat` columns, zero renames, **byte-identical output** (ADR-0015 regression anchor + bit-exact parity tests stay green).

**Delivered:**
- **Consistent feature naming** — one canonical parser/formatter `ut.split_feat_id` / `ut.join_feat_id` replaces ~15 duplicated inline `.split("-")` sites.
- **Standard dataframe structure** — `ut.LIST_COLS_FEAT` (canonical, deterministic column order) applied via `ut.sort_cols_feat` at every output chokepoint (`run`, `run_num`, `get_df_feat`). Tolerant by construction: the dynamic p-value column name, post-hoc `feat_importance`/`feat_impact`, and per-substrate SHAP columns (`feat_impact_'name'`, `mean_dif_'name'`) append after `positions` in stable order and are never dropped — so the order is a lower bound, not a restriction.
- **Documented schema** — the `CPP.run`/`run_num` docstring schema table (explicit dynamic p-value name, tolerant-append note, compositional/positional note) + a consolidated `df_feat` schema entry in `CONTEXT.md`.
[...]

PR #97, opened 2026-06-08
feat: multi-class & regression CPP via label helpers + reference generation (#61)

breimanntools:
Implements #61 as **label transformations + reference synthesis feeding the existing `CPP.run`** — no new `fit_*`/`run_*` methods, no public top-level symbols (no `__init__.py` change).

## What's added (static methods on `SequenceFeature`)
- `get_labels_ovr(labels)` → one-vs-rest binary vectors (X unchanged; loop `run`).
- `get_labels_ovo(labels)` → per class-pair `(row_mask, binary_subset)`.
- `get_labels_quantile(targets, q=0.5)` → continuous → binary via a single quantile cut (no drop).
- `get_df_parts_reference(df_parts, method, n, random_state)` → scrambled / composition-matched reference parts, length-faithful per part by construction.
- `get_df_parts_from_windows(dict_parts)` → assemble a reference `df_parts` from per-part `AAWindowSampler` window sets, enabling **per-part priors** (e.g. coil JMD-N, α-helix TMD, coil JMD-C).

Backend in `_backend/cpp/sequence_feature.py`. Five example notebooks; tests cover golden/error/seed cases plus integration through the real `CPP.run` (#18 `LIST_COLS_FEAT` schema preserved).

## Deferred
- SHAP multi-class/regression → later **pro** PR.
- One-vs-one and **banded** quantile regression need *CPP-on-a-row-subset of `df_parts`* (the `CPP.__init__` binding) — drafted as a separate "part problem" issue.

## Verification (local)
- New suite + full SequenceFeature suite green (225 passed); CPP suite 263 passed; API tests green.
- Docstring checker: 0 defects; doc-signature drift: 0.
- All 5 example notebooks execute clean.

### DIFF
--- a/aaanalysis/feature_engineering/_backend/cpp/sequence_feature.py
+++ b/aaanalysis/feature_engineering/_backend/cpp/sequence_feature.py
@@ -12,9 +12,21 @@ from ._utils_feature_stat import add_stat_
 
 
 # I Helper Functions
+def _part_aa_freq(col=None, aa_list=None, aa_index=None):
+    """Empirical canonical-AA frequency over all strings in one part column."""
+    counts = np.zeros(len(aa_list), dtype=np.int64)
+    for s in col:
+        for c in s:
+            j = aa_index.get(c)
+            if j is not None:
+                counts[j] += 1
+    total = counts.sum()
+    if total == 0:
+        raise ValueError("'df_parts' contains no canonical amino acids for composition matching.")
+    return counts / total
 
 
-# II Main functions
+# II Main Functions
 # Parts and splits
 def get_split_kws_(n_split_min=1, n_split_max=15, steps_pattern=None, n_min=2, n_max=4, len_max=15,
                    steps_periodicpattern=None, split_types=None):
@@ -96,3 +108,66 @@ def get_df_feat_(features=None, df_parts=None, labels=None,
     # Standardize the df_feat column order (issue #18); amino_acids_* append last.
     df = ut.sort_cols_feat(df_feat=df)
     return df
+
+
+# Multi-class / regression label conversion and reference generation
+def get_labels_ovr_(labels=None, label_test=1, label_ref=0):
+    """One-vs-rest: per class, a full-length binary vector (class -> test, rest -> ref)."""
+    labels = np.asarray(labels)
+    classes = sorted(set(labels.tolist()))
+    return {c: np.where(labels == c, label_test, label_ref).astype(int) for c in classes}
+
+
+def get_labels_ovo_(labels=None, label_test=1, label_ref=0):
+    """One-vs-one: per class pair (a, b), a row mask + binary labels for the masked subset."""
+    labels = np.asarray(labels)
+    classes = sorted(set(labels.tolist()))
+    out = {}
+    for i, a in enumerate(classes):
+        for b in classes[i + 1:]:
+            mask = np.isin(labels, [a, b])
+            binary = np.where(labels[mask] == a, label_test, label_ref).astype(int)
+            out[(a, b)] = (mask, binary)
+    return out
+
+
+def get_labels_quantile_(targets=None, q=0.5, label_test=1, label_ref=0):
+    """Single-threshold split of continuous targets into binary labels (>= q-quantile -> test)."""
+    targets = np.asarray(targets, dtype=float)
+    cut = float(np.quantile(targets, q))
+    return np.where(targets >= cut, label_test, label_ref).astype(int)
+
+
+def get_df_parts_reference_(df_parts=None, method=ut.MODE_SCRAMBLED, n=None, rng=None):
+    """Generate reference ``df_parts`` rows, per-part length- and composition-faithful.
+
+    Each generated row borrows the per-part lengths of a randomly chosen real
+    row (so references match the real CPP part lengths by construction).
+    ``scrambled`` shuffles that template part in place (exact per-part
+    composition); ``global_freq`` draws from the column's empirical canonical-AA
+    frequency.
+    """
+    aa_list = list(ut.LIST_CANONICAL_AA)
+    aa_index = {a: i for i, a in enumerate(aa_list)}
+    list_parts = list(df_parts)
+    n_real = len(df_parts)
+    if n is None:
+        n = n_real
+    template_idx = rng.integers(0, n_real, size=n)
+    data = {}
+    for part in list_parts:
+        col = df_parts[part].to_numpy()
+        probs = _part_aa_freq(col=col, aa_list=aa_list, aa_index=aa_index) \
+            if method == ut.MODE_GLOBAL_FREQ else None
+        strings = []
+        for i in range(n):
+            template = col[template_idx[i]]
+            if method == ut.MODE_SCRAMBLED:
+                chars = list(template)
+                rng.shuffle(chars)
+                strings.append("".join(chars))
+            else:
+                strings.append("".join(rng.choice(aa_list, size=len(template), p=probs)))
+        data[part] = strings
+    index = [f"REF{i}" for i in range(n)]
+    return pd.DataFrame(data, index=index)
--- a/aaanalysis/feature_engineering/_sequence_feature.py
+++ b/aaanalysis/feature_engineering/_sequence_feature.py
@@ -1,8 +1,9 @@
 """
 This is a script for the frontend of the SequenceFeature class, a supportive class for the CPP feature engineering.
 """
-from typing import Optional, Union, List
+from typing import Optional, Union, List, Dict, Tuple, Any
 import warnings
+import numpy as np
 import pandas as pd
 
 import aaanalysis.utils as ut
@@ -23,7 +24,9 @@ from ._backend.cpp.utils_feature import (get_df_parts_,
                                          replace_non_canonical_aa_,
                                          get_positions_, get_amino_acids_,
                                          get_df_pos_, get_df_pos_parts_)
-from ._backend.cpp.sequence_feature import (get_split_kws_, get_features_, get_feature_names_, get_df_feat_)
+from ._backend.cpp.sequence_feature import (get_split_kws_, get_features_, get_feature_names_, get_df_feat_,
+                                            get_labels_ovr_, get_labels_ovo_, get_labels_quantile_,
+                                            get_df_parts_reference_)
 from ._backend.cpp_run import _pick_feature_matrix_builder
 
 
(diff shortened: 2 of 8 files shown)


aaanalysis/feature_engineering/_backend/cpp/sequence_feature.py
```diff
@@ -96,3 +108,66 @@ def get_df_feat_(features=None, df_parts=None, labels=None,
...
+
+
+def get_labels_quantile_(targets=None, q=0.5, label_test=1, label_ref=0):
+    """Single-threshold split of continuous targets into binary labels (>= q-quantile -> test)."""
+    targets = np.asarray(targets, dtype=float)
+    cut = float(np.quantile(targets, q))
+    return np.where(targets >= cut, label_test, label_ref).astype(int)
+
+
+def get_df_parts_reference_(df_parts=None, method=ut.MODE_SCRAMBLED, n=None, rng=None):
```
breimanntools, line 141: This capability is already coverd by AAWindowSampler! I need simply a method that consues output from AAWindowSampler (so per part we can generate flexible df_parts! This is the major re-factor

aaanalysis/feature_engineering/_sequence_feature.py
```diff
@@ -887,3 +890,292 @@ def get_df_pos(df_feat: pd.DataFrame = None,
...
+        # Convert labels
+        return get_labels_ovr_(labels=labels, label_test=label_test, label_ref=label_ref)
+
+    @staticmethod
+    def get_labels_ovo(labels: ut.ArrayLike1D = None,
+                       label_test: int = 1,
+                       label_ref: int = 0,
+                       ) -> Dict[Tuple[Any, Any], Tuple[np.ndarray, np.ndarray]]:
+        """
+        Convert multi-class labels into one-vs-one binary label vectors with row masks.
```
breimanntools, line 951: Introduce abbrication ovo (one-vs-one (oVo) in the docstring

aaanalysis/feature_engineering/_sequence_feature.py
```diff
@@ -887,3 +890,292 @@ def get_df_pos(df_feat: pd.DataFrame = None,
...
+        argument of :meth:`CPP.run` / :meth:`CPP.run_num`, enabling regression-style
+        tasks (e.g. thermostability, binding affinity) within the binary CPP
+        framework. A banded top-vs-bottom split (dropping the middle) is out of
+        scope here.
+
+        Parameters
+        ----------
+        targets : array-like, shape (n_samples,)
+            Continuous target values for samples.
+        q : float, default=0.5
```
breimanntools, line 1018: So, I might want a one vs one where I take a max trehsold and then step-wise lowre (so positive are all over 0.8 and negatives are a) allo below or eaul 0.8, b) ... 0.5, c) 0.3 ... To compare different setting but have alwys the same positive. Implement this

aaanalysis/feature_engineering/_sequence_feature.py
```diff
@@ -887,3 +890,292 @@ def get_df_pos(df_feat: pd.DataFrame = None,
...
+
+        Returns
+        -------
+        dict_labels : dict
+            Dictionary mapping each class pair ``(a, b)`` to a tuple
+            ``(row_mask, labels_subset)``, where ``row_mask`` is a boolean array
+            of shape (n_samples,) selecting the pair's samples and
+            ``labels_subset`` is the binary vector over the masked subset.
+
+        Notes
```
breimanntools, line 978: Add for all methdos here a see also CPP class as this will consume the df_parts

aaanalysis/feature_engineering/_sequence_feature.py
```diff
@@ -887,3 +890,292 @@ def get_df_pos(df_feat: pd.DataFrame = None,
...
+        ut.check_number_range(name="q", val=q, min_val=0, max_val=1, exclusive_limits=True, just_int=False)
+        ut.check_number_val(name="label_test", val=label_test, just_int=True)
+        ut.check_number_val(name="label_ref", val=label_ref, just_int=True)
+        if label_test == label_ref:
+            raise ValueError(f"'label_test' ({label_test}) should differ from 'label_ref' ({label_ref}).")
+        # Convert targets
+        return get_labels_quantile_(targets=targets, q=q, label_test=label_test, label_ref=label_ref)
+
+    @staticmethod
+    def get_df_parts_reference(df_parts: pd.DataFrame = None,
```
breimanntools, line 1052: As written before, this class should not do the sampling but consume fro AAWindowSampler. The notebook must be therefore a bit more detailed and make it clar how to use both toghetr. The samples are then matched by position (e.g., AAWindowSampler makes three different snythetic datasets (must match in size, raise warning if this is not the case, perhaps we can provide list_tmd, list_jmd_n, list_jmd_c or so, whatever matches best with the logic of the package) the first sample is simply a combi of fist entry in list_tmd, list_jmd ...

aaanalysis/feature_engineering/_sequence_feature.py
```diff
@@ -887,3 +890,292 @@ def get_df_pos(df_feat: pd.DataFrame = None,
...
+        ut.check_str_options(name="method", val=method,
+                             list_str_options=[ut.MODE_SCRAMBLED, ut.MODE_GLOBAL_FREQ])
+        ut.check_number_range(name="n", val=n, min_val=1, accept_none=True, just_int=True)
+        random_state = ut.check_random_state(random_state=random_state)
+        # Generate reference parts
+        rng = np.random.default_rng(random_state)
+        return get_df_parts_reference_(df_parts=df_parts, method=method, n=n, rng=rng)
+
+    @staticmethod
+    def get_df_parts_from_windows(dict_parts: Dict[str, Union[pd.DataFrame, ut.ArrayLike1D]] = None,
```
breimanntools, line 1111: this is acutally the one matching to what I wanted. The get_df_refernces can then most likely be removed, but let us discuss it. The ideas from the comment before can be intergred if not alredy applied here

tests/unit/sequence_feature_tests/test_sf_labels_reference.py
breimanntools, line 1: This does not follow my testing template!! Use positive and negtive tests, use property-based testing with hyptohesis and do simple and complex test classes. I do not want to see such tests again. Each parameter must be tested proerply. Update the context.md and the uster guide (where testing hygine is described) rember this and update your memory

breimanntools (commented):
Comprehensive Review: Multi-Class & Regression CPP via Label Helpers + Reference Generation
Your Self-Review Comments (1-7) + What Was Missed
Comment 1: AAWindowSampler Integration ✓

You correctly identified that get_df_parts_from_windows() should consume AAWindowSampler output (not re-implement sampling). The implementation is correct, but:

What was missed:

    Minimal notebook example — expand sf_get_df_parts_from_windows.ipynb to show the full AAWindowSampler → assembly workflow with biological motivation (coil JMD-N, α-helix TMD, coil JMD-C)
    No guidance on input format optimization (DataFrame vs. list-of-strings)
    Error message when window counts mismatch could suggest recovery steps (e.g., "Ensure all window lists are in the same sampling order")

Comment 2: OvO Abbreviation ✓

Correctly flagged. The abbreviation "OvO" appears inline but is not formally introduced in a Notes section.

What was missed:

    Add formal definition to get_labels_ovo Notes: "One-vs-one (OvO): K classes → K(K-1)/2 class pairs; each pair requires its own row subset of df_parts and CPP instance."
    Consistency: standardize across all docstrings (OvR vs. one-vs-rest, OvO vs. one-vs-one)

Comment 3: Banded Quantile Regression ✓

Correctly deferred, but the feature request is not fully captured.

What was missed:

    Create a dedicated GitHub issue (#XX) with the specific use case: "Tiered threshold exploration with fixed positive set" (e.g., positive = all ≥0.8; negatives = ≤0.8, ≤0.5, ≤0.3 in separate runs)
    Clarify in docstring: "A banded top-vs-bottom split (dropping middle samples) and tiered thresholds are out of scope here."
    Add pseudocode comment showing how users can currently implement it: for threshold in [0.8, 0.5, 0.3]: labels = (targets >= threshold).astype(int)

Comment 4: Missing "See Also" Sections ✓✓✓ CRITICAL

Completely absent from all 6 methods. Each label method and reference method should reference :class:CPP and each other.

Required additions:
Code

See Also
--------
* :class:`aaanalysis.CPP` — the feature engineering class that consumes these label vectors and reference parts.
* :meth:`SequenceFeature.get_labels_ovr` — alternative for K-class problems (if not self-reference).
* :meth:`SequenceFeature.get_df_parts_reference` — generating synthetic negative class when none exists.

Comment 5: Notebook Clarity & AAWindowSampler Integration ✓

The notebook is too minimal (3 cells, no context).

What was missed:

    Add biological motivation: Why would you use separate generators for JMD vs. TMD? (Physicochemical priors differ)
    Expand with explicit check: assert len(dict_parts['jmd_n']) == len(dict_parts['tmd']) == len(dict_parts['jmd_c']), "Window counts must match"
    Show what happens if counts mismatch (demonstrate the error and how to fix it)
    Document the "matching by position" contract: "The i-th window from each part forms the i-th reference row. Order must be consistent across all parts."

Comment 6: Reference Generation Design ✓

You noted potential redundancy between get_df_parts_reference_() and get_df_references. This needs clarification:

What was missed:

    Is get_df_references a separate method that should be removed? Or is this a future enhancement?
    Add docstring note: "This method generates synthetic backgrounds when no natural negative class exists. It differs from directly downsampling or bootstrapping the positive class."
    Document when to use scrambled vs. global_freq:
        scrambled: Preserves exact per-part composition; safer if negative class should be compositionally similar
        global_freq: More diverse; empirical amino acid frequency may not match real proteins

Comment 7: Testing Hygiene ✓✓✓ CRITICAL

Tests do NOT follow your template. Current violations:

What was missed:

    ✗ NO property-based testing with hypothesis — only hardcoded examples
    ✗ NO per-parameter systematic testing — label_test and label_ref tested together, not independently
    ✗ NO positive/negative test class separation
    ✗ NO "simple → complex" test progression
    ✗ Missing edge cases:
        get_labels_quantile: constant targets (zero variance), tied quantile values, extreme q (0.001, 0.999)
        get_df_parts_reference: very small n_real (n=2-3, random template pool quality), non-canonical AAs (X, *)
        All methods: single-element input arrays, empty inputs (should raise)
    ✗ NO update to context.md or user guide with this test file as the example of proper testing hygiene

Action: Rewrite test_sf_labels_reference.py following your testing template (golden/error/seed cases, hypothesis strategies, parametrized per-param tests). Update docs with this file as reference.
Additional Comments (8-21): Further Improvements
8. Numerical Stability & Type Validation 🔴 HIGH

Location: Backend functions (get_labels_ovr_, get_labels_ovo_, get_labels_quantile_)

Issue:

    get_labels_quantile_ converts targets to float silently — if inputs are float32, precision loss near quantile boundary is possible with tied values
    get_labels_ovr_ and get_labels_ovo_ use sorted(set(labels.tolist())) which assumes labels are hashable & orderable. Float labels like [0, 0.0, 1, 1.5] could behave unexpectedly

Suggestion:

    Add validation: labels must be integer-type or integer-like (reject floats, strings)
    Document: "Labels must be hashable and orderable; typically integers (0, 1, 2, ...). String labels or mixed types may produce unexpected ordering."
    Add comment to get_labels_quantile_: "Note: targets are converted to float64 for numerical stability. Very small variance targets (<1e-10) should be pre-scaled by the user."

9. Composition Bias in Global Frequency Mode 🟡 MEDIUM

Location: get_df_parts_reference_() with method='global_freq'

Issue: The _part_aa_freq() function pools amino acids from the entire column. If part lengths vary widely (some jmd_n=10, some jmd_c=20), longer parts dominate the frequency distribution, biasing shorter parts toward over-represented residues.

Suggestion:

    Document: "The global_freq method computes empirical amino acid frequency from the entire column. If part lengths vary significantly, longer parts may dominate the frequency; use scrambled for more conservative reference generation."
    Add optional parameter for per-length stratification (future enhancement, defer now)
    Add warning if >5% of characters are non-canonical (currently skipped silently): if skipped_count / total_count > 0.05: warn("...")

10. Missing Label Type Validation 🔴 HIGH

Location: Frontend: check_labels() utility; Backend: get_labels_ovr_, get_labels_ovo_

Issue: The validator doesn't explicitly reject float or string labels, only ensures they match label_test/label_ref. But backend assumes integer-like ordering from sorted(set(...)). This can cause subtle bugs with float labels like [0.0, 0.5, 1.0].

Suggestion:

    Add explicit validation in frontend:
    Python

    if not all(isinstance(label, (int, np.integer)) for label in labels):
        raise ValueError("'labels' must be integer-type; received mixed or float labels.")

    Add docstring note: "Labels must be integers. For continuous targets, use get_labels_quantile() to discretize first."

11. Quantile Edge Cases & Pre-Validation 🔴 HIGH

Location: get_labels_quantile() frontend

Issue:

    If targets have zero variance (all constant), np.quantile() succeeds but returns the single value; then >= comparison yields all True or all False. The docstring mentions CPP will reject this, but it's passive (relies on downstream error).
    If targets are extremely imbalanced at the quantile (e.g., q=0.95 splits 95/5), statistical power for the minority class is low.

Suggestion: Add pre-validation:
Python

# After converting targets to float array:
if len(np.unique(targets)) == 1:
    raise ValueError("Targets have no variance (all equal to {targets[0]:.2f}); "
                     "quantile split will produce only one class.")

cut = float(np.quantile(targets, q))
n_test = (targets >= cut).sum()
n_ref = (targets < cut).sum()
if min(n_test, n_ref) / len(targets) < 0.1:
    warnings.warn(f"Quantile q={q} produces imbalanced split ({n_test}/{n_ref}); "
                  "consider adjusting q for better class balance.")

Document in Notes:

    "Targets are assumed continuous; ordinal data may produce unexpected splits."
    "Extreme quantile choices (q<0.1 or q>0.9) may yield statistically underpowered groups."

12. Window Ordering & Alignment Clarity 🟡 MEDIUM

Location: get_df_parts_from_windows() docstring

Issue: The docstring says "Rows are aligned by position" but doesn't define "position" in context of AAWindowSampler. If windows are not in the same order across parts, or if one list is accidentally shuffled, the biological meaning of assembled rows breaks silently.

Suggestion: Add explicit note to Parameters & Notes:
Code

Parameters
----------
dict_parts : dict
    ... (existing) ...
    **Important:** All window lists must be in the same order across parts. 
    If using AAWindowSampler.sample_synthetic, use identical n and random_state 
    for deterministic ordering, or manually sort all lists identically.

Notes
-----
* Rows are aligned by position: the i-th window from each part forms the i-th 
  reference row. Ensure all window lists share the same ordering.
* If windows are in different orders (e.g., one sorted alphabetically, one random), 
  the assembled df_parts will not represent biologically sensible sequences.

13. Computational Complexity Not Documented 🟡 MEDIUM

Location: All docstrings

Issue: Users can't predict performance for large K (number of classes) or N (samples). No guidance on when OvO becomes prohibitively expensive (k² CPP instances).

Suggestion: Add Complexity note to each method:
Code

Notes
-----
* **Complexity - get_labels_ovr:** O(n_samples × n_classes). Scales linearly in both dimensions; 
  minimal overhead for K=10, K=100, etc.
* **Complexity - get_labels_ovo:** O(n_samples × n_classes²). For K=10 → 45 pairs; 
  K=20 → 190 pairs; each pair requires its own CPP instance. 
  **Recommend OvO only for K<~10; use OvR for larger problems.**
* **Complexity - get_labels_quantile:** O(n_samples log n_samples) due to np.quantile. 
  Typically negligible compared to CPP runtime.
* **Complexity - get_df_parts_reference (scrambled):** O(n × n_parts × L) where L=avg 
  part length. Highly efficient; ~0.1ms for typical inputs.
* **Complexity - get_df_parts_reference (global_freq):** Same as scrambled, but includes 
  an additional O(total_length) pass to compute frequencies.

14. OvO Requires CPP Instance Rebuild (Undocumented Workflow Burden) 🔴 HIGH

Location: get_labels_ovo() docstring

Issue: The docstring mentions "running CPP on a row-subset without rebuilding the instance is a separate enhancement", but this is a MAJOR practical burden that users will struggle with. Users must:

    Call get_labels_ovo(labels) → returns O(k²) pairs
    For each (a, b) pair:
        Apply row_mask to subset df_parts
        Create a new CPP instance on the subset
        Call CPP.run()
        Concatenate results manually

This means k² CPP instantiations for k classes.

Suggestion: Add full workflow example to docstring:
Code

Examples
--------
.. code-block:: python

    # Multi-step workflow for OvO:
    df_seq = aa.load_dataset(name="DOM_GSEC", n=10)
    sf = aa.SequenceFeature()
    df_parts = sf.get_df_parts(df_seq=df_seq)
    multiclass = np.array([0, 0, 0, 1, 1, 1, 2, 2, 2, 2])
    
    frames = []
    for (a, b), (row_mask, binary) in aa.SequenceFeature.get_labels_ovo(multiclass).items():
        df_pair = df_parts[row_mask]  # Subset rows for this pair
        cpp = aa.CPP(df_parts=df_pair)  # Create NEW instance for subset
        df_feat = cpp.run(labels=binary, n_filter=3)
        df_feat['pair'] = f"({a},{b})"
        frames.append(df_feat)
    
    df_feat_ovo = pd.concat(frames, ignore_index=True)

Future enhancement: Consider a helper method SequenceFeature.run_ovo_series(df_parts, labels, split_kws, ...) that encapsulates this loop and returns concatenated results directly.
15. Notebook Examples: Missing Failure Cases 🔴 HIGH

Location: All 5 notebooks

Issue: Each notebook shows only the "happy path". Users don't learn the validation boundaries or how to recover from errors.

Suggestion: Add a final cell to each notebook: "What can go wrong?" section demonstrating validation errors:

sf_get_labels_ovr.ipynb:
Python

# Error: single-class labels
try:
    aa.SequenceFeature.get_labels_ovr([1, 1, 1])
except ValueError as e:
    print(f"Error: {e}")

# Error: equal test/ref labels
try:
    aa.SequenceFeature.get_labels_ovr([0, 1], label_test=1, label_ref=1)
except ValueError as e:
    print(f"Error: {e}")

sf_get_labels_quantile.ipynb:
Python

# Constant targets (edge case)
try:
    aa.SequenceFeature.get_labels_quantile([5.0, 5.0, 5.0], q=0.5)
except ValueError as e:
    print(f"Error: {e}")

# Imbalanced split (warning case)
targets = [0, 0, 0, 0, 0, 0.01]
labels = aa.SequenceFeature.get_labels_quantile(targets, q=0.99)
print(f"Split: {labels.sum()} positives, {len(labels) - labels.sum()} negatives")

sf_get_df_parts_from_windows.ipynb:
Python

# Error: mismatched window counts
try:
    aa.SequenceFeature.get_df_parts_from_windows({
        'jmd_n': [window1],
        'tmd': [window2, window3]  # 2 instead of 1
    })
except ValueError as e:
    print(f"Error: {e}")

# Error: invalid part name
try:
    aa.SequenceFeature.get_df_parts_from_windows({'not_a_part': [...]})
except ValueError as e:
    print(f"Error: {e}")

sf_get_df_parts_reference.ipynb:
Python

# Behavior with non-canonical AAs
df_parts_with_X = pd.DataFrame({
    'jmd_n': ['AAAX', 'CCCC'],
    'tmd': ['EEEEX', 'FFFFF'],
    'jmd_c': ['HHHHX', 'KKKK']
})
df_ref = aa.SequenceFeature.get_df_parts_reference(df_parts_with_X, method='global_freq')
# Show how X is skipped/handled

This teaches users the error boundaries and recovery strategies.
16. Type Hints: Return Types Could Be More Explicit 🟢 LOW

Location: Method signatures

Suggestion: Clarify ambiguous return type annotations:

    get_labels_ovr() → Dict[int, np.ndarray]: Add docstring note that keys are sorted class labels
    get_labels_ovo() → Dict[Tuple[Any, Any], ...]: Specify Tuple[int, int] instead of Tuple[Any, Any] to clarify classes are hashable integers
    get_df_parts_from_windows(dict_parts: Dict[str, Union[pd.DataFrame, ut.ArrayLike1D]]): Clarify ArrayLike1D means Sequence[str] for window strings, not generic array

17. Missing Integration Test: Imbalanced Multi-Class 🟡 MEDIUM

Location: Test file (add to integration section)

Issue: No tests for imbalanced class distributions, which are common in real data.

Suggestion: Add test class:
Python

class TestImbalancedMultiClass:
    """Test label helpers with unequal class sizes."""
    
    def test_get_labels_ovr_imbalanced(self):
        labels = [0, 0, 0, 0, 0, 0, 1, 1, 2]  # 6:2:1 split
        d = SF.get_labels_ovr(labels)
        assert d[0].sum() == 6
        assert d[1].sum() == 2
        assert d[2].sum() == 1
    
    def test_get_labels_ovo_imbalanced_pair(self):
        labels = [0, 0, 0, 0, 0, 1, 2, 2, 2]  # Class 0 >> Class 1
        o = SF.get_labels_ovo(labels)
        mask_01, binary_01 = o[(0, 1)]
        assert mask_01.sum() == 6  # 5 zeros + 1 one
        assert binary_01.sum() == 1  # One sample of class 0 (test)
    
    def test_quantile_imbalanced_split(self):
        targets = np.array([0, 0, 0, 0, 0, 0, 10])  # Extreme outlier
        labels = SF.get_labels_quantile(targets, q=0.5)
        # Median = 0, so 6 samples = 0 (ref), 1 sample = 10 (test)
        assert labels.sum() == 1  # Very imbalanced

Document: "These methods work with imbalanced class distributions. However, downstream CPP feature statistics (AUC, p-value) may be statistically underpowered for minority classes. Consider pre-filtering or resampling if class imbalance is extreme (>100:1)."
18. Reproducibility: RNG State & Multiple Calls 🟡 MEDIUM

Location: get_df_parts_reference() docstring

Issue: If users call get_df_parts_reference() multiple times with the same random_state, they get identical results (good). But if they increment the seed each time (seed 42, 43, 44), the internal RNG state differs because each call consumes randomness. This is non-obvious.

Suggestion: Add explicit note to docstring:
Code

Notes
-----
* Reproducibility: Calling with the same random_state produces identical output 
  (deterministic). To generate multiple *independent* reference sets:
  
  Option A (deterministic but identical):
    ref1 = SequenceFeature.get_df_parts_reference(df_parts, random_state=42)
    ref2 = SequenceFeature.get_df_parts_reference(df_parts, random_state=42)  # Same as ref1

  Option B (independent):
    ref1 = SequenceFeature.get_df_parts_reference(df_parts, random_state=42)
    ref2 = SequenceFeature.get_df_parts_reference(df_parts, random_state=43)  # Different
    
* The RNG state progresses through each call; using seed+1, seed+2, etc. provides 
  pseudo-independence. For true statistical independence, consider deriving from 
  a master RNG: `rng = np.random.default_rng(master_seed); seeds = rng.integers(0, 2**31, 3)`

19. Docstring Consistency: Terminology 🟢 LOW

Location: All docstrings

Issue: Terminology varies: "binary vector", "binary label vector", "binary label array". Should be consistent.

Suggestion: Standardize across all docstrings to: "binary label array" (since np.ndarray is returned)

Add consistent sentence to all label methods:
Code

Notes
-----
* The returned array is directly usable as the ``labels`` argument to :meth:`CPP.run`.

20. Performance: Optimize sorted(set(...).tolist()) → np.unique() 🟡 MEDIUM

Location: get_labels_ovr_() line 117, get_labels_ovo_() line 124

Issue: Current code:
Python

classes = sorted(set(labels.tolist()))  # O(n log n) due to conversion + sorting

Better:
Python

classes = np.unique(labels)  # O(n log n) but optimized, ~5x faster in practice

Suggestion:
Python

# Optimization comment:
def get_labels_ovr_(labels=None, label_test=1, label_ref=0):
    """One-vs-rest: per class, a full-length binary vector (class -> test, rest -> ref)."""
    labels = np.asarray(labels)
    # np.unique is faster than sorted(set(...).tolist()); returns sorted array automatically
    classes = np.unique(labels)
    return {c: np.where(labels == c, label_test, label_ref).astype(int) for c in classes}

Benchmark: For labels.shape=(10,000,):

    Current: ~0.5ms
    np.unique: ~0.1ms
    5x speedup!

Apply same to get_labels_ovo_().
21. Usage Guide: Decision Tree for Method Selection 🔴 HIGH

Location: New section in SequenceFeature class docstring or PR description

Issue: Users don't know which method to use for their problem (K-class? Regression? Need negative class?).

Suggestion: Add a "Method Selection Guide" to the SequenceFeature class docstring (Notes section):
Code

**When to use which label/reference method:**

Problem: I have multi-class labels (0, 1, 2, ...)
├─ Sub-goal: I want one feature set per class
│  └─> Use get_labels_ovr()
│      Pros: Simple, reuses full df_parts, loop in user code
│      Cons: Must manually loop and concatenate results
│
├─ Sub-goal: I want pairwise comparisons (A vs B, A vs C, ...)
│  └─> Use get_labels_ovo()
│      Pros: Captures pair-specific discriminative features
│      Cons: O(k²) class pairs; each requires CPP instance rebuild; complex loop

Problem: I have continuous targets (thermostability, binding affinity, ...)
├─ Sub-goal: Convert to binary classification for CPP
│  └─> Use get_labels_quantile(q=0.5)  [or other percentile]
│      Pros: Bridges regression → binary CPP framework
│      Cons: Quantile threshold is arbitrary; no statistical test guides choice

Problem: I don't have a natural negative/reference class
├─ Sub-goal: Generate synthetic background
│  └─> Use get_df_parts_reference(method='scrambled')
│      Pros: Exact composition preservation, biologically conservative
│      Cons: Scrambled may not reflect true negative class distribution
│
├─ Sub-goal: Generate synthetic background with per-part biological priors
│  └─> Use get_df_parts_from_windows(dict_parts)
│      Pros: Each part (JMD-N, TMD, JMD-C) can have own physicochemical prior
│      Cons: Requires AAWindowSampler expertise; more setup

**Quick decision matrix:**
K-class multi-class?     → get_labels_ovr() (simple) or get_labels_ovo() (pairwise)
Continuous target?       → get_labels_quantile()
No negative class?       → get_df_parts_reference() (simple) or 
                            get_df_parts_from_windows() (biologically motivated)

Summary of Action Items
🔴 CRITICAL (Block merge until fixed):

    Add "See Also" sections to all 6 methods linking to CPP and each other
    Rewrite test file with hypothesis-based tests, parametrized per-parameter tests, positive/negative splits, edge cases
    Upgrade docstrings with per-parameter validation rules, complexity analysis, and failure case examples
    Fix OvO workflow burden documentation with full example code

🟡 HIGH PRIORITY (Should fix before merge):

    Add label type validation (integer-only)
    Add quantile edge case pre-validation (constant targets, imbalance warning)
    Expand notebooks with biological motivation & error cases
    Add computational complexity notes to all methods

🟢 NICE-TO-HAVE (Next iteration or follow-up PR):

    Optimize sorted(set().tolist()) → np.unique()
    Add reproducibility guidance (RNG state management)
    Add usage guide / decision tree to class docstring
    Document composition bias in global_freq mode
    Add imbalanced class tests

PR #110, opened 2026-06-10
feat: model-free feature pruning (variance + correlation) — issue #32

### LINKED ISSUE #32: Add feature selection utilities for AAanalysis outputs
## Goal
Enable selection of informative features.

## Tasks
- [ ] Implement variance filtering
- [ ] Add correlation-based filtering
- [ ] Integrate with pipeline

## Acceptance criteria
- Users can reduce feature space easily

breimanntools:
Closes #32 (narrowed scope: variance + correlation filtering only; model-based selection is `TreeModel.select_features`, ADR-0023, already shipped).

## What

Adds two model-free, post-hoc **feature pruning** methods on `SequenceFeature`, returning a row-filtered `df_feat` (same schema) that composes **before** `TreeModel.select_features`:

- **`prune_by_variance(df_feat, df_parts, threshold=0.0)`** — drops near-constant features (column variance over all samples `<= threshold`). Constant columns are detected by zero peak-to-peak range, so `threshold=0` removes **exactly** the constant features (robust to float epsilon).
- **`prune_by_correlation(df_feat, df_parts, max_cor=0.7)`** — drops empirically redundant features (`|corr| > max_cor` over the actual samples), keeping the higher-`abs_auc` feature; deterministic via `[abs_auc, abs_mean_dif]` ordering. Reuses the `NumericalFeature.filter_correlation` primitive.

Both build the feature matrix via `SequenceFeature.feature_matrix` or accept a precomputed `X` (covering `run_num` tables / build-once reuse).

## Key design point

`CPP.run` already redundancy-reduces internally, but on **scale-vector correlation** (`df_scales.corr()`) + position overlap — independent of any dataset. `prune_by_correlation` is **empirical** (sample-level), catching dataset-specific redundancy the in-run filter cannot. The two are complementary by design (ADR-0026).

## Acceptance criteria (issue #32 KPIs)

- ✅ Variance filter at threshold 0 removes exactly the zero-variance columns (golden test).
- ✅ Correlation filter at `t` guarantees no retained pair has `|corr| > t` (property test) and is byte-identical across runs (determinism test).
- ✅ Output passes the `df_feat` schema check (#18); ≥1 unit + property tests; example notebooks run.

## Changes

- Backend helper `_backend/num_feat/filter_variance.py::filter_variance_`.
- `SequenceFeature.prune_by_variance` / `prune_by_correlation` (numpydoc, named Returns, per-method Examples includes).
- 32 tests (unit + hypothesis property + determinism + golden), house hygiene template.
- Two example notebooks under `examples/feature_engineering/` (executed outputs).
- `CONTEXT.md` feature-pruning glossary; **ADR-0026**.

No new public symbols (additive methods on an exported class) → outside CONFIRM-FIRST; no new deps.

### DIFF
--- a/aaanalysis/feature_engineering/_sequence_feature.py
+++ b/aaanalysis/feature_engineering/_sequence_feature.py
@@ -704,12 +704,18 @@ class SequenceFeature:
             DataFrame of scales with letters typically representing amino acids. Default from :meth:`load_scales`
             unless specified in ``options['df_scales']``.
         threshold : float, default=0.0
-            Features with column variance ``<= threshold`` are dropped. With the default ``0.0`` only
-            strictly constant features are removed; raise it to prune low-variance features more aggressively.
+            Minimum **population variance** (``numpy`` ``var``, ``ddof=0``) a feature's column must exceed
+            to be kept; the threshold is in **variance units**, not standard-deviation units. Feature
+            values are means of (typically ``[0, 1]``-normalized) scale values, so variances are small —
+            commonly below ``0.1`` — and a useful range is: ``0.0`` removes only strictly constant
+            features, while ``~0.01`` to ``~0.05`` also prunes low-variance features. The variance is
+            computed **over all provided samples** (every row of ``df_parts`` / ``X``, both classes
+            together) per feature column — not the test group only, and not per split.
         X : array-like, shape (n_samples, n_features), optional
-            Pre-computed feature matrix aligned column-for-column with ``df_feat``. If given, it is used
-            directly and ``df_parts`` / ``df_scales`` are ignored (e.g. to reuse a matrix or to prune a
-            :meth:`CPP.run_num` ``df_feat``).
+            Pre-computed feature matrix. Column ``i`` must correspond to the feature in row ``i`` of
+            ``df_feat`` (same order). If given, it is used directly and ``df_parts`` / ``df_scales`` are
+            ignored (e.g. to reuse a matrix across pruning calls or to prune a :meth:`CPP.run_num`
+            ``df_feat``).
         accept_gaps : bool, default=False
             Whether to accept missing values by enabling omitting for computations (if ``True``).
         n_jobs : int, None, or -1, default=1
@@ -723,6 +729,13 @@ class SequenceFeature:
 
         Notes
         -----
+        * **Variance metric**: population variance (``ddof=0``) of each feature column over all samples.
+          A feature that is constant across the samples (zero peak-to-peak range) is treated as exactly
+          zero variance, so ``threshold=0.0`` removes precisely the constant features even when floating
+          point would otherwise leave a tiny non-zero variance.
+        * **Scope**: variance reflects how much a feature varies across *your* samples; it is unrelated to
+          CPP's in-run pre-filter, which screens *candidate* features by the **test-group** standard
+          deviation (``max_std_test``) rather than the spread over all samples.
         * Recommended pruning order: variance (this method) -> correlation
           (:meth:`SequenceFeature.prune_by_correlation`) -> :meth:`TreeModel.select_features`.
         * A pruning that retains no feature (e.g. ``threshold`` above every feature's variance) raises a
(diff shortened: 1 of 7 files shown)


aaanalysis/feature_engineering/_sequence_feature.py
```diff
@@ -653,6 +669,195 @@ def feature_matrix(self,
...
+        ----------
+        df_feat : pd.DataFrame, shape (n_features, n_feature_info)
+            Feature DataFrame with a unique identifier, scale information, statistics, and positions for
+            each feature.
+        df_parts : pd.DataFrame, shape (n_samples, n_parts), optional
+            DataFrame with sequence parts. Used to build the feature matrix; not required if ``X`` is given.
+        df_scales : pd.DataFrame, shape (n_letters, n_scales), optional
+            DataFrame of scales with letters typically representing amino acids. Default from :meth:`load_scales`
+            unless specified in ``options['df_scales']``.
+        threshold : float, default=0.0
```
breimanntools, line 706: What kind of variance (std, var) please also provide a value range to get a better idea of the threshold. why should we remove features with 0 variance, 0 variance in the test. Or over the whole dataset? Adresse it in the docs Here a bit more points "Or ultra-concise (one sentence each):

    Variance type: Confirm this is population variance and whether the threshold is in variance or std units.
    Threshold range: Document typical thresholds (e.g., 0.0 for constants, 0.01–0.1 for low-variance pruning).
    Scope: Explicitly state zero-variance detection is computed over all provided samples, not per-split.
"

aaanalysis/feature_engineering/_sequence_feature.py
```diff
@@ -653,6 +669,195 @@ def feature_matrix(self,
...
+        else:
+            X = check_match_df_feat_X(df_feat=df_feat, X=X)
+        # Prune features whose variance is at or below the threshold
+        is_selected = filter_variance_(X, threshold=threshold)
+        if not is_selected.any():
+            raise ValueError(f"'threshold' ({threshold}) removed all features. Lower it to retain features.")
+        df_feat = df_feat[is_selected].reset_index(drop=True)
+        return df_feat
+
+    def prune_by_correlation(self,
```
breimanntools, line 770: How is this different from our NumericalFeature.filter_correlation method. Make he differnce clear if it exist. Which feature is discareded in case of correlation? (I would nomaly sort them by auc_abs and the one with lower auc_abs is removed, this should be clear. But the user can sor the as he wishes. The X must match if the have differnt order. This must be in

breimanntools (commented):
Both methods have solid implementations (tests confirm correctness), but the documentation needs to be more explicit about:

    Variance: metric type, units, threshold ranges, scope (all samples vs. per-split)
    Correlation: how it differs from existing primitives, deterministic tie-breaking logic, X-matrix alignment requirements

These clarifications will prevent user confusion and misus

breimanntools:
Addressed the documentation feedback in 36954294 (docstrings only — implementations unchanged, tests still green).

**`prune_by_variance`**
- **Variance type / units:** now states it's **population variance** (`numpy var`, `ddof=0`), and that `threshold` is in **variance units, not std**.
- **Threshold range:** documented — feature values are means of (typically `[0,1]`) scale values, so variances are small (often `< 0.1`); `0.0` removes only constants, `~0.01–0.05` also prunes low-variance features.
- **Scope:** explicit that variance is over **all provided samples (both classes), per feature column** — *not* the test group and *not* per-split; contrasted with CPP's in-run `max_std_test` (test-group std) pre-filter.

**`prune_by_correlation`**
- **Difference from `NumericalFeature.filter_correlation`:** added — that primitive is `X → boolean mask` keeping the **first** column of a correlated pair (input order); this method is `df_feat`-in/out, builds `X` for you, and **ranks by `abs_auc` first** so the weaker feature is dropped.
- **Tie-break:** explicit that the **lower-`abs_auc`** feature of each over-threshold pair is removed, **independent of input row order**, and the result is `abs_auc`-descending and deterministic.
- **X alignment:** documented that a supplied `X` must have **column `i` = feature in row `i`** of `df_feat`; the method reorders both together, so a mis-aligned `X` would correlate the wrong features.

PR #111, opened 2026-06-10
Add parameter-coverage meta-test + branch-coverage gate (#84)

### LINKED ISSUE #84: Add parameter-coverage + branch-coverage checks
## Problem

CI measures **line coverage only**: `test_coverage.yml` runs
`pytest tests -m "not regression" --cov=aaanalysis --cov-fail-under=88` with no
`--cov-branch`, so both arms of a conditional count as covered when only one is
exercised (the `check_cat` filter in #77 is a live example of an untested
branch hiding under green line coverage). There is also **no check that every
public parameter is actually exercised** by a test — a parameter can be added to
a public signature, documented, and never called with a non-default value while
coverage stays green. These two gaps let real behavioural holes pass CI.

## Goal

Add branch-coverage measurement with a gate and a parameter-coverage check for
public signatures, so untested branches and unexercised public parameters fail
CI — staying within the unit-test scope (integration/e2e remain v2-deferred).

## Requirements

- [ ] Enable branch coverage: add `--cov-branch` to the `test_coverage.yml`
      pytest invocation (and document running it locally). **Editing
      `.github/workflows/test_coverage.yml` is CONFIRM-FIRST — ask before
      changing CI.**
- [ ] Establish the current branch-coverage baseline and set a
      `--cov-fail-under` (or a branch-specific gate) at-or-just-below baseline so
      it ratchets, mirroring the existing line gate of 88 (ADR-0016 honest
      package-only coverage).
- [ ] Add a parameter-coverage check (small script under `dev_scripts/` or a
      `tests/` meta-test): introspect each public symbol in `aaanalysis.__all__`,
      enumerate its non-`self` parameters, and assert each is referenced by name
      in at least one test call site (or an explicit allowlist of intentionally
      untested params with reasons).
- [ ] Record both numbers (branch %, param-coverage %) so future PRs see drift.

## KPIs / Acceptance criteria

- [ ] CI reports a **branch-coverage percentage** and fails if it drops below the
      committed gate (state the baseline number, e.g. "branch ≥ X%").
- [ ] The parameter-coverage check reports a number and **fails when a new public
      parameter has 0 test references**; target **≥ 95%** of public parameters
      covered (remainder on a justified allowlist).
- [ ] Both checks run in the existing `pytest -c tests/pytest.ini` / coverage
      workflow with no new required heavy dependency.

## Scope / non-goals

- Unit-test tooling only. **Integration / e2e / system tests stay v2-deferred**
  per `.claude/rules/sharp-edges.md` — explicitly out of scope here.

## Dependencies

- Split from #41 (Testing, non-deferred items). Overlaps with #77 (which closes
  one specific uncovered branch). Relates ADR-0016.

## Standards checklist

- [x] CONFIRM-FIRST: `.github/workflows/test_coverage.yml`
- [x] tests/meta-tests; coverage measured package-only (`--cov=aaanalysis`)
- [x] no `print()` in any new script (use `ut.print_out` if library-side)

breimanntools:
Closes #84.

## What

Adds the two coverage checks #84 asks for, in ADR-0016's package-only, ratcheted spirit.

### 1. Parameter coverage (`tests/unit/api_tests/test_param_coverage.py`)
Enumerates every public parameter of every symbol in `aaanalysis.__all__` (`__init__` + every public method; properties, `self`/`cls`, `*args/**kwargs` excluded) and **fails if a parameter is neither referenced as a keyword argument anywhere under `tests/` nor on an explicit `ALLOWLIST` with a reason.**

- Detection is **global and name-based** (a param is covered if its name is used as a kwarg at any call site) — simple and robust to test-layout drift, with a documented false-positive surface for ambient names. The branch gate below catches the code paths those names mask.
- **Pro symbols** that are `missing_feature_stub` lambdas (extra absent) are skipped with a recorded reason, so the check is green in a core-only `[dev]` env and enforces the pro surface only where `[pro]` is installed.
- Machinery guard-tests prevent the gate from passing vacuously (broken enumeration/parsing).

Current: **979 public params, test-covered 97.6%, 0 uncovered, 23 allowlisted** (visual-only passthroughs: `plot_legend`, `AAlogoPlot` logomaker kwargs; plus the `AAMut`/`SeqMut` whole-class gap — see notes).

### 2. Branch coverage
`test_coverage.yml` adds `--cov-branch` and **drops `--cov-fail-under`** (once branch is on it gates the *combined* number, silently redefining the line floor). A post-pytest step `.github/scripts/check_branch_coverage.py` parses `coverage.xml` and enforces an **independent line floor (>=88) and branch floor (>=80)**.

Measured baseline: **line 92.68%, branch 85.56%.** Gates set conservatively at-or-below baseline (ADR-0016 ratchet); raise in tracked steps.

### 3. Closing the behavioural gaps
Real kwargs added at existing call sites: `AAclust.fit` `on_center`/`merge`, `AAclust.name_clusters` `shorten_names`, `SequenceFeature` `targets`/`dict_parts`, `dPULearn.compare_sets_negatives` `remove_non_neg`, `CPPPlot.update_seq_size` `max_x_dist`, `plot_settings` `show_options`, `AnnotationPreprocessor.register_feature` `subcategory`/`normalization`, `ShapModel.eval` `shap_values` (NotImplementedError contract).

### 4. Docs
ADR-0016 addendum (D4/D5) records the non-obvious trade-offs; `.claude/rules/workflow.md` gets the local commands.

## Notes for review
- `AAMut`/`SeqMut` have **no dedicated unit-test suite** at all; the global check only surfaces their uniquely-named `from_aa`/`to_aa`/`show_delta`, which are allowlisted with a reason pointing at that whole-class gap. Recommend a follow-up issue to stand up those suites.
- The branch gate (80) is deliberately conservative vs the 85.56% baseline to absorb the `[dev]`-only CI env (pro excluded) difference; ratchet up once CI reports its number.

### DIFF
--- a/.claude/rules/workflow.md
+++ b/.claude/rules/workflow.md
@@ -18,6 +18,16 @@ pytest tests/unit/aaclust_tests/test_aaclust_fit.py::TestAAclustFit::test_valid_
 # coverage (matches CI gate: test_coverage.yml uses --cov-fail-under=88)
 pytest --cov=aaanalysis --cov-fail-under=88 tests
 
+# branch coverage + the branch/line gate (issue #84). --cov-branch reports a
+# branch-rate; the helper parses coverage.xml and fails below the committed
+# line/branch gates (do NOT use --cov-fail-under with --cov-branch: it would
+# gate the combined number and silently redefine the line floor).
+pytest tests -m "not regression" --cov=aaanalysis --cov-branch --cov-report=xml -n auto -c tests/pytest.ini
+python .github/scripts/check_branch_coverage.py
+
+# parameter coverage meta-test (fast, no coverage run needed)
+pytest tests/unit/api_tests/test_param_coverage.py -x -vv -c tests/pytest.ini
+
 # run notebooks — LOCAL gate only (nbmake is NOT in blocking CI; re-run + re-commit
 # outputs before every push; RTD renders committed outputs but does not execute them)
 pytest --nbmake --nbmake-timeout=120 tutorials/ examples/
--- a/.github/scripts/check_branch_coverage.py
+++ b/.github/scripts/check_branch_coverage.py
@@ -1,0 +1,75 @@
+"""
+This is a script for the CI branch- and line-coverage gate (issue #84).
+
+It parses the cobertura ``coverage.xml`` produced by
+``pytest --cov-branch --cov-report=xml`` and exits non-zero if either the
+line-rate or the branch-rate falls below its committed gate. It runs as a step
+*after* the coverage pytest invocation in
+``.github/workflows/test_coverage.yml`` because ``coverage.xml`` is only written
+at the end of the pytest session. It lives under ``.github/scripts/`` (not
+``dev_scripts/``, which is git-ignored) so the CI checkout can run it.
+
+Branch coverage cannot be gated with ``pytest --cov-fail-under`` directly:
+once ``--cov-branch`` is on, that flag checks the *combined* line+branch number,
+which would silently re-define the line gate. Parsing the two rates separately
+keeps an honest, independent line floor and branch floor (ADR-0016 D4).
+
+This is CI tooling, not library code, so it prints to stdout for the CI log
+(the package itself never calls ``print`` — it uses ``ut.print_out``).
+
+Local use::
+
+    pytest tests -m "not regression" --cov=aaanalysis --cov-branch \\
+        --cov-report=xml -n auto -c tests/pytest.ini
+    python .github/scripts/check_branch_coverage.py
+"""
+import sys
+import xml.etree.ElementTree as ET
+from pathlib import Path
+
+# Committed gates in percent. Ratcheted at-or-just-below the measured baseline
+# (ADR-0016); raise only once the suite clears the next step. The line gate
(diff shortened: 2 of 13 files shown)


breimanntools:
### CI status: all blocking checks green ✅

- Full test matrix (ubuntu + windows × py 3.10–3.14): **pass**
- Test Coverage job incl. the new line/branch gate: **pass** — CI reported `line 95.36% (gate 88%) | branch 88.50% (gate 80%)`
- CodeQL / code-quality / Read the Docs: **pass**
- Parameter-coverage meta-test: **965 public params, 23 allowlisted, 0 uncovered**

### Note on `codecov/project` (red, non-blocking)
This is the **expected one-time methodology change** from enabling `--cov-branch`: Codecov now counts *partial* branches, so the combined metric dips vs the base commit's line-only number, and `project: target: auto` (no-regression) trips on the first branch-enabled commit. It's the same class of intentional one-time drop as ADR-0016 D1 (`--cov=./` → `--cov=aaanalysis`). `master` is unprotected so it does not block merge; once this lands, the new base carries branch data and subsequent PRs compare apples-to-apples. No action needed unless you'd prefer to set the codecov `project` status `informational` during the climb.

Issue #43, opened 2026-04-11
AAclust


breimanntools:
Clarify AAclust applicability to protein-level clustering and representative selection

Goal:
Make explicit that AAclust is not limited to clustering AA scales, but can be applied to clustering proteins based on feature representations (e.g. CPP features).

Documentation Improvements

- [ ] Clarify that AAclust operates on feature matrices (not restricted to scales)
- [ ] Explicitly state applicability to protein sequences via CPP feature vectors
- [ ] Add example: clustering proteins based on CPP features

Representative Selection

- [ ] Document that medoid corresponds to a representative protein (real sequence)
- [ ] Clarify biological interpretation of representative sequences
- [ ] Add example: selecting representative proteins per cluster
- [ ] Score quality of representative (distance to cluster members)
- [ ] Support returning top-k representatives per cluster

Use Case Extension

- [ ] Add tutorial section: "Clustering proteins and selecting representatives"
- [ ] Demonstrate workflow: sequence → CPP features → AAclust → representative proteins

Outcome:
AAclust is clearly positioned as a method for clustering proteins and selecting representative sequences based on interpretable features.

breimanntools:
**Scope note from external feedback triage** (see `docs/feedback_antibody_triage.md`).

Two small AAclust ergonomics surfaced by the antibody/nanobody project:
- `random_state` — **already shipped** (`AAclust.__init__`, threaded to the sklearn model; clustering is deterministic when set). Their "non-determinism" observation predates this.
- `n_clusters >= n_unique_samples` currently **raises** (`check_match_X_n_clusters`) rather than clamping; and `n_clusters` semantics (exact-k vs upper-bound) confused them ("AAclust-30 ≈ AAclust-20"). Worth, under this issue: a graceful clamp option and a docstring that states the exact-k semantics explicitly.

No new issue; folding these into #43.

Issue #22, opened 2026-04-09
CPP + Embedding integration


breimanntools:
## Goal  
Integrate embedding-based representations (e.g. PLMs) with CPP to enable joint analysis and comparison of sequence-derived and learned feature spaces.

### Input & interface

- [ ] Define standardized embedding input format (aligned with df_seq; JSON / matrix)
- [ ] Ensure consistent indexing (sequence, residue, position)

### Integration into pipeline

- [ ] Integrate embeddings as additional feature space alongside CPP
- [ ] Enable joint processing of CPP features and embeddings
- [ ] Ensure compatibility with existing feature pipeline and outputs

### Comparison & analysis

- [ ] Compute similarity in embedding space (cosine / distance)
- [ ] Compare embedding similarity vs CPP similarity
- [ ] Compute correlations between CPP features and embeddings
- [ ] Perform clustering in embedding space and compare with CPP-based clustering

### Representation mapping

- [ ] Map embedding dimensions to CPP feature space (scale-based interpretation)
- [ ] Cluster embedding dimensions and relate to AAontology categories
- [ ] Identify embedding-derived feature groups with physicochemical meaning

### Evaluation

- [ ] Test integration on example datasets
- [ ] Evaluate consistency between CPP and embedding-based representations
- [ ] Identify cases of agreement vs divergence

## Acceptance criteria  

- Embeddings can be used as input alongside CPP features  
- CPP and embedding representations can be compared systematically  
- Embedding features can be partially interpreted via CPP / AAontology  
- Integration works within the existing pipeline without breaking compatibility

breimanntools:
**Scope note from external feedback triage** (antibody/nanobody non-specificity project; see `docs/feedback_antibody_triage.md`).

The per-*residue* embedding-fusion path is already shipped — `EmbeddingPreprocessor` + `combine_dict_nums` + `CPP.run_num` (struct/embedding/fused arms) handle `{entry: (L, D)}` tensors. What that project still had to build outside AAanalysis, and what remains the live slice of this issue:

- **Per-sequence pooled-vector fusion**: concatenating a *one-vector-per-sequence* block (e.g. mean-pooled ESM-1v) with CPP features. `combine_dict_nums` operates on per-residue tensors, so a pooled `(n, D)` block isn't a first-class input.
[...]

PR #145, opened 2026-06-12
feat(embed): EmbeddingPreprocessor.fetch_embeddings + pool_embeddings ([embed] extra, ADR-0029)

breimanntools:
## What
In-package protein-language-model embeddings — the `fetch_*` sibling of `fetch_alphafold` / `fetch_uniprot` (verb taxonomy ADR-0017, extended by **ADR-0029**).

- **`EmbeddingPreprocessor.fetch_embeddings(df_seq, mode="protein"|"residue", model=...)`** — downloads a curated PLM (ESM-2 ×5, ESM-1b, ProtT5, ProstT5) from the Hugging Face Hub and computes embeddings. `mode="protein"` → `(n, D)` matrix (mean/max/cls pooling) for `AAclust.select_proteins` / a classifier; `mode="residue"` → `{entry: (L, D)}` raw arrays that feed `encode` → `CPP.run_num`. Hardware-aware size guard with a smaller-model recommendation; `on_failure`/`return_df` mirror the `fetch_*` siblings.
- **`EmbeddingPreprocessor.pool_embeddings(embeddings, pooling=)`** — reduce per-residue arrays to per-protein vectors (the simple statistical counterpart to CPP's pooling).

## Design (ADR-0029)
- **Compute-first**: no clean per-accession UniProt embedding endpoint exists (only bulk per-proteome HDF5), so `source="uniprot"` is reserved/deferred; `source="compute"` ships.
- Embeddings returned **raw** — normalization stays `encode`'s job (ADR-0011).
- Class stays in **core**; heavy deps lazy-imported in-method (afragmenter/msms precedent), so the rest of `EmbeddingPreprocessor` works on a base install.

## CONFIRM-FIRST surfaces (pre-approved by maintainer)
- **`pyproject.toml`** — new **`[embed]`** extra (`torch`, `transformers`, `sentencepiece`, `huggingface_hub`), isolated from `[pro]` to avoid ~50× its install weight.
- **`aaanalysis/__init__.py`** — `_EXTRA_MODULES["embed"]` + install map (no `__all__` change; methods on an already-exported class).

## Testing
- **65 tests green**: backend foundation (registry / hardware / recommend / pooling), frontend validation, fake-loader compute logic (`on_failure`, pooling, truncation), + a real `esm2_t6_8M` integration test.
- **CI-safe**: the real-model test is `skipif`-no-torch, so it skips where `[embed]` is absent; the rest are networkless. Verified end-to-end with the real model locally.
- Two executed example notebooks (`ep_fetch_embeddings`, `ep_pool_embeddings`). Note: `nbmake` of `ep_fetch_embeddings` needs `[embed]` locally (nbmake is not blocking CI; RTD renders committed outputs).

## Relates
- Relates #22 (CPP + embedding integration); enables the embedding input for the data-representations tutorial (follow-up to #43).
- No issue auto-closed.


breimanntools (commented):
Looks all good to push

aaanalysis/data_handling/_embed_preproc.py
```diff
@@ -430,3 +437,223 @@ def build_cat(
             metric=metric,
         )
         return df_cat
+
+    def fetch_embeddings(
+        self,
+        df_seq: pd.DataFrame = None,
+        mode: Literal["protein", "residue"] = "protein",
+        model: str = "esm2_t12_35M",
```
breimanntools: Please Liteal with different options that are valid

aaanalysis/data_handling/_embed_preproc.py
```diff
@@ -430,3 +437,223 @@ def build_cat(
...
+        df_seq : pd.DataFrame, shape (n_samples, n_seq_info)
+            DataFrame containing an ``entry`` column with unique protein identifiers and
+            a ``sequence`` column with full protein sequences. Output rows are aligned to
+            ``df_seq``.
+        mode : {'protein', 'residue'}, default='protein'
+            ``'protein'`` returns one pooled vector per protein; ``'residue'`` returns
+            the per-residue ``(L, D)`` array per protein (feeds :meth:`encode`).
+        model : str, default='esm2_t12_35M'
+            Registry key of the PLM to use (e.g. ``'esm2_t6_8M'``, ``'esm2_t33_650M'``,
+            ``'prott5_xl_u50'``, ``'prostt5'``).
```
breimanntools: Please give here all key examples or should we allow simple any? We need an informative error message which are availibe. Please make an overview table comparing the 5-10 once (like size, performance, ....)

aaanalysis/data_handling/_embed_preproc.py
```diff
@@ -430,3 +437,223 @@ def build_cat(
...
+            ``np.ndarray`` of shape ``(n_samples, D)`` row-aligned to ``df_seq``
+            (``mode='protein'``), or ``{entry: (L, D)}`` of raw per-residue arrays
+            (``mode='residue'``).
+        df_seq_out : pd.DataFrame
+            Returned only when ``return_df=True``: an echo of ``df_seq`` plus a boolean
+            ``embeddings_ok`` column.
+
+        Notes
+        -----
+        * Embedding extraction is deterministic (eval mode), so no ``random_state`` /
```
breimanntools, line 610: Here the table should be for comparing the differnce PLMs

breimanntools (commented):
Elaborate on the different PLMs and give size, pro, con or so (brief overview of top 5-10). Also add the restriction of what can run local or what can not run local. We can fetch per protein from UniProt. This can also be describe (when should we do this and when should we compute ourself and pool with aanalsis)?

breimanntools:
Addressed the review comments in `2f8d4308`:

- **`model` as `Literal`** (comment at `_embed_preproc.py:445`): the `model` parameter is now a `Literal` of the 8 valid registry keys, so IDEs autocomplete and type-check it. Runtime validation still lists the valid options on an unknown key (informative `ValueError`).
- **List all keys + informative error** (`:484`): the `model` docstring now names every key (`esm2_t6_8M` … `prostt5`) and states that an unknown key raises a `ValueError` listing the valid options.
- **PLM comparison table** (`:484`, `:523`): added a *Notes* table comparing all 8 models — params, embedding dim, ~RAM (CPU), whether each runs on a typical laptop CPU (Local? = yes/slow/GPU), and best-for. Plus the local-vs-GPU restriction (the size guard warns and suggests a smaller model when a model won't fit).
- **UniProt vs. compute-locally**: added a *Notes* paragraph on when to fetch UniProt's precomputed ProtT5 per-protein embeddings (large, fully-covered Swiss-Prot sets on CPU) vs. compute here (novel/mutated sequences, non-ProtT5 models, or per-residue output for `encode` → `CPP.run_num`). `source='uniprot'` remains reserved for a future release (no clean per-accession endpoint exists yet — see ADR-0029).

Docstring checker clean, 46 embedding tests + param-coverage green, no doc-signature drift.

PR #152, opened 2026-06-13
docs(api): surface implicit contracts at the docstring surface (#135)

### LINKED ISSUE #135: Docstring discoverability: surface implicit contracts at the API surface
## Problem
Several load-bearing contracts live only in `CONTEXT.md` (the domain glossary), not in the numpydoc docstrings users actually read:
- Numerical mode requires `NumericalFeature.get_parts(df_seq, dict_num) → dict_num_parts` **before** `CPP.run_num` — not stated in `CPP.run_num`'s docstring (a user calling `run_num(dict_num=…)` gets a `dict_num_parts` error).
- `dict_num` must be `[0,1]`-normalized (CPP's `max_std_test` is calibrated for it) — only in `CONTEXT.md`.
- `df_seq` has 5 formats; consumers (e.g. `SeqMut`) don't say which they require.
- `[pro]` gating is invisible in class docstrings — users discover it via `ImportError`.

## Goal
Elevate these contracts to the relevant docstrings (+ a `[pro]` marker in pro class summaries) — **non-breaking docs**.

## Requirements
- [ ] `CPP.run_num` + `NumericalFeature.get_parts`: a `Notes` block stating the `get_parts → run_num` order and the `[0,1]` contract (and what breaks if unnormalized).
- [ ] Pro classes/functions (`ShapModel`, `StructurePreprocessor`, `AnnotationPreprocessor`, `comp_seq_sim`, `filter_seq`, `scan_motif`): a `[pro]` marker / "requires `aaanalysis[pro]`" line in the class/function summary.
- [ ] `SeqMut` and other `df_seq` consumers: reference the required `df_seq` format (cross-link to the `SequenceFeature.get_df_parts` format docs).
- [ ] Run the docstrings checker (`.claude/skills/docstrings/scripts/check_docstrings.py`).

## KPIs / Acceptance
- [ ] Each named docstring contains its contract (grep/checker test); a reader can determine pro-gating + normalization + call-ordering without running code.

## Scope / non-goals
Docs only; no API/behavior change.

## Dependencies
Part of the **API ergonomics** epic #126.

## Standards checklist
- [ ] numpydoc (named Returns, Examples) · docstrings checker green · no code change to `[pro]` gating.

breimanntools:
Closes #135

## What

Elevates four load-bearing contracts that previously lived only in `CONTEXT.md` to the numpydoc docstrings users actually read. **Docs only — no API or behavior change.**

1. **Call order + `[0, 1]` contract.** `CPP.run_num` and `NumericalFeature.get_parts` Notes now state the two-step `get_parts → run_num` order and the `[0, 1]` normalization contract — *including what breaks if `dict_num` is unnormalized*: the `max_std_test` pre-filter is silently miscalibrated and the feature funnel keeps/drops the wrong features (no error raised).
2. **`[pro]` markers.** `ShapModel`, `StructurePreprocessor`, `AnnotationPreprocessor`, `comp_seq_sim`, `filter_seq`, `scan_motif` now carry a `[pro]` / "requires `aaanalysis[pro]`" marker in their summary, so gating is visible without first hitting an `ImportError`.
3. **`df_seq` format cross-link.** `SeqMut.mutate` / `scan` / `suggest` cross-link the canonical `df_seq` format spec (`SequenceFeature.get_df_parts`). (`comp_seq_sim` / `filter_seq` / `scan_motif` already spell out their exact columns inline, so they were left unchanged.)
4. **Guard test.** New `tests/unit/api_tests/test_docstring_contracts.py` — an `ast`-based meta-test (env-independent; holds even without the `pro` extras installed) codifying the issue KPI so these contracts cannot silently drift back out of the docstrings.

Plus a Release Notes entry under *v1.1.0 (Unreleased) → Changed*.

## Local gates (all green)

- `check_docstrings.py`: **0 defects** (6 advisories, all pre-existing / identical to baseline).
- `doc_signature_drift.py`: **0 drift**.
- `tests/unit/api_tests/`: **62 passed** (incl. the new 11-case guard).
- Package imports cleanly; all cross-references resolve.

### DIFF
--- a/aaanalysis/data_handling_pro/_annot_preproc.py
+++ b/aaanalysis/data_handling_pro/_annot_preproc.py
@@ -92,8 +92,8 @@ def _check_scores_unit_range(scores, name="score"):
 # II Main Functions
 class AnnotationPreprocessor:
     """
-    Preprocessing class for per-residue post-translational modification (PTM) /
-    functional-site annotations.
+    Preprocessing class (**[pro]**, requires ``aaanalysis[pro]``) for per-residue
+    post-translational modification (PTM) / functional-site annotations.
 
     Collects per-residue annotations — fetched from UniProt
     (:meth:`fetch_uniprot`) or ingested from a user / predictor table
--- a/aaanalysis/data_handling_pro/_struct_preproc.py
+++ b/aaanalysis/data_handling_pro/_struct_preproc.py
@@ -348,7 +348,8 @@ def _run_get_dssp_internal(df_seq, pdb_folder, features, ss_mode,
 # II Main Functions
 class StructurePreprocessor:
     """
-    Preprocessing class for protein structure features (PDB / CIF / AlphaFold).
+    Preprocessing class (**[pro]**, requires ``aaanalysis[pro]``) for protein
+    structure features (PDB / CIF / AlphaFold).
 
     Turns local structure files into the ``[0, 1]``-normalized per-residue
     ``dict_num`` consumed by :meth:`CPP.run_num`. Each ``encode_*`` method
--- a/aaanalysis/explainable_ai_pro/_shap_model.py
+++ b/aaanalysis/explainable_ai_pro/_shap_model.py
@@ -238,8 +238,9 @@ def check_match_df_feat_shap_values(df_feat=None, shap_values=None, drop=False, 
 # II Main Functions
 class ShapModel:
     """
-    SHAP Model class: A wrapper for SHAP (SHapley Additive exPlanations) [Lundberg20]_ explainers to obtain
-    Monte Carlo estimates for feature impact [Breimann25a]_.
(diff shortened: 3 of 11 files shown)


breimanntools (commented):
Yes. This is very good. Can we make an autoatmic scan whether all the pro modules are actually clearly shown in the docstring, making sure that the contract is always valid.

breimanntools:
Addressed the review (commit 8dd6871e): the docstring guard now **auto-discovers** every public symbol of every `aaanalysis/*_pro/` subpackage — it reads each `__init__.py` `__all__` and follows the `from ._module import Name` line to the definition, then asserts each carries the `[pro]` install marker in its summary. So a newly added pro feature is covered with no edit to the test and fails until it advertises gating. A `test_pro_discovery_is_not_vacuous` floor guards against discovery silently finding nothing. All 63 api-tests pass locally.

PR #156, opened 2026-06-13
docs(tutorials): CPP data-representations tutorial (scales / embeddings / structure → feature_map)

breimanntools:
## What (PR 2 — the data-representations tutorial)
New Feature Engineering tutorial `tutorial3d_data_representations.ipynb` (placed after the CPP intro, `tutorial3c_cpp`). It shows the point you asked for: **scales, embeddings, and structure all feed the same CPP machinery and provide different kinds of features**, with a **`CPPPlot.feature_map` for each** so the difference is visible.

- **Arm 1 — scales** → `CPP.run` (AAontology categories).
- **Arm 2 — PLM embeddings** → `EmbeddingPreprocessor.fetch_embeddings`/`encode` → `CPP.run_num` (learned channels).
- **Arm 3 — structure/DSSP** → `StructurePreprocessor.fetch_alphafold` + `encode_dssp` → `CPP.run_num` (structural channels).

Same `df_feat` schema and map layout each time; only the value source changes. Closes the loop that any of these representations feeds the same downstream tools (`CPPPlot`, `TreeModel`, `AAclust.select_proteins`).

## Notes
- Uses a model-free importance (normalized `abs_auc`) for the map (there's no public per-protein feature matrix for the numerical arm to fit a tree on; noted inline, with a pointer to `TreeModel.add_feat_importance`).
- Runs end-to-end with the real ESM-2 model + live AlphaFold fetch (now reliable after #151), `n_jobs=1`; committed with the three feature maps.
- Re-running needs `aaanalysis[embed]` + `aaanalysis[pro]` + the CD-HIT/DSSP tools; `nbmake` is a local-only gate, RTD renders the committed outputs.
- Wired into the Feature Engineering toctree.


breimanntools (commented):
Looks good! But can we make first sequence, then structure and then embeddings! Also the annotation shouldbe introduced after embeddings!

breimanntools:
Addressed in `cc78a9d7`: reordered the arms to **sequence → structure → embeddings**, and added a fourth **annotation** arm after embeddings (`AnnotationPreprocessor.fetch_uniprot` + `encode` → `CPP.run_num` over UniProt residue features: phospho / glyco / disulfide / binding / active-site / mod-res). Now four `CPPPlot.feature_map`s, all re-executed live with fresh outputs.

PR #160, opened 2026-06-13
perf(cpp): vectorize get_df_parts_ row driver

### LINKED ISSUE #19: Optimize performance of df_parts computation
## Problem
df_parts computation is a performance bottleneck.

## Goal
Improve efficiency of feature extraction.

## Tasks
- [ ] Identify inefficient operations
- [ ] Optimize loops/vectorization
- [ ] Benchmark before/after
- [ ] Ensure no functional regression

## Acceptance criteria
- Significant speed improvement
- Same output as before

breimanntools:
## Summary

`SequenceFeature.get_df_parts` was bottlenecked by a per-row
`DataFrame.apply(axis=1)` in the backend `get_df_parts_`. Each row paid for a
pandas `Series` materialization, while the actual per-row work (`create_parts` +
`get_dict_part_seq`) is cheap.

This replaces the `apply` driver with a list comprehension over the raw column
arrays (`zip` of `.to_numpy()` columns), branching once on `pos_based` instead of
per row. The per-row primitives are **unchanged**, so the output `df_parts` is
**byte-identical** (parts, column order, index, dtypes, values).

Closes #19.

## Performance

10k-row position-based `df_seq` (local, macOS):

| call | before | after | speedup |
|---|---|---|---|
| `all_parts=True` | ~669 ms | ~140 ms | ~4.8× |
| `list_parts="tmd"` | ~678 ms | ~134 ms | ~5.1× |

Per row: 67 µs → 14 µs. (The cost was independent of the number of parts
requested, confirming the `apply` driver — not part-building — was the
bottleneck.) Also drops the now-unneeded `df_seq.copy()` + temporary `'parts'`
column.

### Why not Cython?
The repo already ships a Cython kernel for the *feature matrix* (the numeric
inner loop). `df_parts` is string-slicing assembled into a DataFrame of Python
`str` objects — that object allocation is irreducible, so Cython's ceiling here
is low, and after this fix `df_parts` is no longer the pipeline bottleneck. The
pure-Python change captures the real win (removing per-row `Series`
materialization) with zero new infrastructure.

## Parity verification
Captured `get_df_parts` output on unmodified `master` across all input formats
(position-based, part-based, seq-TMD, anchor, gapped/short-terminus, custom jmd
lengths, single + `all_parts`) and asserted `old.equals(new)` post-change:
**10/10 byte-identical** (values, index, dtypes).

## Tests / gates (local)
- Full fast unit suite (`-m "not regression"`): **4112 passed, 7 skipped** (py3.13 + py3.14).
- `sequence_feature_tests` + `cpp_tests` (not regression): **760 passed**.
- Param-coverage meta-test: 5/5.
- Exact-value CPP regression anchor: self-skips off the canonical cell
  (ADR-0015; runs nightly on Linux/py3.11).
- New golden test `test_gap_padding_golden` pins the `-`-padded composite parts
  on a short sequence — the trickiest path the vectorized driver must preserve.

## Ripple
No public-API/signature change, so no docstring/example/cheat-sheet/tables/API-ref
churn. Touches only: the backend driver, one new golden test, and a release-notes
bullet under `v1.1.0 (Unreleased)`.

### DIFF
--- a/aaanalysis/feature_engineering/_backend/cpp/utils_feature.py
+++ b/aaanalysis/feature_engineering/_backend/cpp/utils_feature.py
@@ -121,13 +121,9 @@ def _get_df_pos_long(df=None, col_cat="category", col_val=None):
 
 
 # Get df parts
-def _extract_parts(row, jmd_n_len, jmd_c_len, pos_based, list_parts):
-    """Helper function to extract parts from a single row."""
-    if jmd_c_len is not None and jmd_n_len is not None and pos_based:
-        seq, tmd_start, tmd_stop = row[ut.COLS_SEQ_POS]
-        jmd_n, tmd, jmd_c = create_parts(seq, tmd_start, tmd_stop, jmd_n_len, jmd_c_len)
-    else:
-        jmd_n, tmd, jmd_c = row[ut.COLS_SEQ_PARTS]
+def _extract_parts(parts, list_parts):
+    """Build the {part: seq} dict for one sequence from its (jmd_n, tmd, jmd_c) base parts."""
+    jmd_n, tmd, jmd_c = parts
     # Get dictionary of parts and filter by list_parts
     dict_part_seq = ut.get_dict_part_seq(tmd=tmd, jmd_n=jmd_n, jmd_c=jmd_c)
     return {part: dict_part_seq[part] for part in list_parts}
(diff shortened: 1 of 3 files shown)


breimanntools (commented):
Looks good

PR #169, opened 2026-06-13
perf: same-output speedups for AAWindowSampler filters, AAclust medoid distances, dPULearn KLD

breimanntools:
## Summary

Four **same-output** performance improvements, each validated against the original implementation *before* inclusion (byte/array-identical or `np.allclose` atol=1e-10) and benchmarked. **No public API change; default behavior preserved.**

| Hotspot | Output check | Speedup (validated, at scale) |
|---|---|---|
| `AAWindowSampler` `filter_redundancy` (greedy, buffer-view) | identical (integer-ratio decisions) | ~33x |
| `AAWindowSampler` `filter_similarity_to_test` (vectorized) | identical | ~35x (300 test windows) |
| `AAclust` `_dist_to_medoids` correlation branch (einsum) | identical (max\|Δ\|=0 on fixtures) | ~46x (60k samples) |
| `dPULearn.eval` per-feature KLD (`kullback_leibler_divergence_`) | bit-exact | ~2.2x (`comp_kld=True`) |

## How it was validated

These came out of a whole-library performance audit. Every candidate was measured against its original implementation before shipping; **only changes that are both same-output and actually faster are included.** Notably, several audit candidates were **dropped after benchmarking proved they don't help**:

- **FASTA writer** (`iterrows`→vectorized): ~1.0x even at 100k rows — it's I/O + string-format bound, not `iterrows`.
- **TreeModel CV parallelization**: ~1.0x at fine *and* coarse grain — joblib spawn / IPC overhead cancels it in this regime.
- **load_scales / load_dataset caching** and **dPULearn AUC**: already cached / already parallel — no headroom.

The reliable lever here is **vectorizing Python loops** (no IPC tax), which is what the shipped window-filter and medoid-distance changes do; identity filtering is an exact integer ratio so keep/drop decisions are bit-identical, with a scalar fallback for ragged-length inputs.

## Notes

- `kullback_leibler_divergence_` gains an `n_jobs` argument (private; default `1` → public `comp_kld` stays serial/unchanged). `dPULearn.eval` threads it through, so KLD now parallelizes by default — consistent with its AUC, which already did. Output is bit-exact.
- New equivalence tests pin each change to a reference implementation; full unit suite green (4112 passed).


breimanntools:
**Automated review pass (before manual review):**

- `/code-review` (high effort): one finding — `_encode_equal_length` used `.encode("latin-1")`, which would raise `UnicodeEncodeError` on a non-ASCII residue where the original `window_identity` handled any character. Fixed in 44775823 (catch → fall back to the exact scalar path; added a non-latin-1 equivalence test). All other candidates verified as parity (einsum NaN matches `np.corrcoef`; `np.array_split` can't receive ≤0 since `check_n_jobs` normalizes and the serial branch handles `n_jobs==1`) or by-design (public `comp_kld` stays serial).
- `/security-review`: no findings — changes are in-process numerical math (einsum, joblib over trusted internal arrays, numpy comparisons); no new injection/deserialization/path/network surface.

Equivalence is locked by committed tests (window filters vs scalar reference incl. ragged + non-latin-1 fallback; `_dist_to_medoids` vs per-pair `corrcoef`; KLD parallel-vs-serial). Local: full fast unit suite green pre-merge (4112), affected suites green post-merge (584).

breimanntools:
Skipping manual review — automated gates green (/code-review finding fixed, /security-review clean) and the equivalence + full local suites pass. Arming auto-merge (squash); it will merge once the CI matrix is green.

PR #172, opened 2026-06-13
perf: vectorize AAWindowSampler candidate-center filter + motif PWM scan (same output)

breimanntools:
## Summary

Batch 3 of the performance effort — **same-output** vectorization of hot `AAWindowSampler` loops. Each change is verified against a verbatim copy of the original (exact equality) and benchmarked in isolation. **No public API change.**

| Hotspot | Output | Speedup (isolated) |
|---|---|---|
| `candidate_centers_` band filter | identical (integer L1 distances) | **~44x** (2.2s→0.05s, 200 calls × 200 excludes) |
| `_scan_protein_` PWM scoring (`sample_motif_matched`) | identical hits + scores (left-to-right sum) | **~12x** |
| `score_window_pwm_` aa-index hoist | identical scores | ~1.5x |

## Method (per the batch-1 discipline)

Benchmarked candidates first: `candidate_centers_` (2.2s/200 calls) and `_scan_protein_` were genuinely hot; `_compute_centers` / `_position_aa_freq` were *not* (one-time, not in any hot path) and were left alone. Each shipped change is bit-identical:
- `candidate_centers_`: nearest-excluded distance via `np.abs(centers[:,None]-excl).min(axis=1)`; integer distances → identical band keep/drop.
- `_scan_protein_`: per-window score summed over positions left-to-right via an index matrix; non-canonical residues add `0.0` (bit-identical to skipping). Hit order and float scores match.
- `score_window_pwm_`: hoist the per-call aa-index dict to a module constant.

Equivalence tests (`test_batch3_vectorization_equiv.py`) cover non-canonical residues, `allowed_centers`, band-filter edges, and early returns. Full fast unit suite green (4235 passed).

Note: in-process back-to-back timing is unreliable here (a heavy Python loop contaminates the next measurement), so the speedups above are from isolated runs; correctness is pinned by exact-equality tests regardless.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,9 @@ and a suite of site-localization metrics and plotting helpers.
   `AAWindowSampler` redundancy/similarity filtering (vectorized, ~30x at scale),
   `AAclust` sample-to-medoid correlation distances (single pass), and per-feature
   Kullback-Leibler divergence used by `dPULearn.eval(comp_kld=True)`
-  (parallelized, honors `options['n_jobs']`).
+  (parallelized, honors `options['n_jobs']`). Plus `AAWindowSampler`
+  candidate-center band filtering (~40x) and `sample_motif_matched` PWM scoring
+  (~12x), vectorized with identical output.
 
 ### Deprecated
 - None. The strict-semver deprecation policy and the `deprecated` decorator are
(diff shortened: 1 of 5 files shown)


breimanntools:
**Automated review pass:** `/code-review` (high, two independent finders + edge-case testing) → no findings; verified `.tolist()` yields Python int/float matching all call sites, `+0.0` is bit-identical to skipping non-canonical residues, hit order and index bounds correct. `/security-review` → no findings (pure numpy/dict math, no new attack surface). Equivalence pinned by `test_batch3_vectorization_equiv.py`; full unit suite green (4235). Shipping on green.

PR #174, opened 2026-06-13
perf: vectorize encode_one_hot (same output)

breimanntools:
## Summary

Finishes Batch 3: vectorizes `SequencePreprocessor.encode_one_hot`. The per-residue loop (which rebuilt the alphabet→index dict for *every* residue, then built a list of arrays and flattened per sequence) is replaced with a char-matrix → index-matrix → scatter. **Byte-identical output**, ~3x faster.

- Same `int64` dtype, same position-major-then-alphabet column order (matches `list_features` and the original flatten), gaps → all-zero blocks.
- The frontend (`check_match_list_seq_alphabet`) already guarantees every residue is in `alphabet + gap`, so the original's `KeyError`-on-unknown path is unreachable — vectorizing changes nothing observable.
- Verified identical (matrix + features + dtype) across `pad_at` C/N and custom alphabets.

`encode_integer` was benchmarked too (0.009s / 5000 seqs — already fast, not a hot path) and **left unchanged** to avoid a no-op change.

Equivalence test added (`TestEncodeOneHotEquivalence`) pinning the output to a per-residue reference. Full fast unit suite green (4243 passed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,8 @@ and a suite of site-localization metrics and plotting helpers.
   Kullback-Leibler divergence used by `dPULearn.eval(comp_kld=True)`
   (parallelized, honors `options['n_jobs']`). Plus `AAWindowSampler`
   candidate-center band filtering (~40x) and `sample_motif_matched` PWM scoring
-  (~12x), vectorized with identical output.
+  (~12x), and `SequencePreprocessor.encode_one_hot` (~3x), vectorized with
+  identical output.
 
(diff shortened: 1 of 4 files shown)


breimanntools:
Same-output (byte-identical) vectorization, verified vs a per-residue reference across pad_at C/N + custom alphabets; KeyError path proven unreachable via frontend validation. Full unit suite green (4243). Shipping on green.

breimanntools:
CI green across the matrix; merging.

Issue #186, opened 2026-06-13
perf: Batch 6 — remaining same-output vectorization candidates (whole-library audit tail)


breimanntools:
## Problem
The library-wide same-output performance sweep merged 9 optimizations (#169, #172, #174, #175) and tracks two more strands in #180 (encode_pdb alignment cache + pro-IO fetch). A **tail of smaller same-output candidates** from the original whole-library audit remains untouched — individually modest, collectively worthwhile, and all low-risk because they preserve output.

## Goal
Work this candidate queue opportunistically, each as a small focused PR with an equivalence test, under the established discipline: **benchmark every candidate in isolation BEFORE claiming a win** (the audit's % estimates proved unreliable — FASTA, TreeModel CV, encode_pae, etc. were no-ops; only ship what measurably helps and is byte/`allclose`-identical).

### Structure (`[pro]`, data_handling_pro/_backend/struct_preproc)
- [ ] `encode_disulfide`: O(n²) SG–SG distance loop → vectorized (same formula as the merged contact-count fix). *High confidence.*
- [ ] `encode_pdb` per-entry coordinate/residue cache: extract CA/CB coords + ordered residue list once and share across the 13 encoders (the audit's larger struct envelope; the merged alignment cache #180-A captures the alignment part). *Med — more invasive; needs golden tests on P1/P2/AF_TINY fixtures.*
- [ ] `_plddt_per_residue`: compute once, reuse across the 3 pLDDT encoders.
- [ ] DSSP `align_dssp_full` / `run_dssp_full`: reuse one session-scoped `PairwiseAligner` across entries (own copy, separate from encode_pdb).

### Sequence analysis
- [ ] `comp_seq_cons` (`[pro]`): per-position `Counter`+log → numpy `log2` (`allclose`). *High confidence.*
- [ ] `AAlogo.get_df_logo_info`: cache the information-content matrix (currently a redundant logomaker call).
- [ ] `get_df_logo_` / `_retrieve_tmd_aligned`: per-row `apply` padding → `pd.Series.str.pad`.

### Feature engineering
- [ ] `SequenceFeature.prune_by_correlation` / `filter_correlation_`: `corrcoef` + nested triangle → `triu_indices` + early-exit.
- [ ] CPP `filtering` redundancy (`_simplify.py`): per-candidate set ops → one vectorized overlap matrix.
- [ ] CPP `_eligible_candidates_`: cache `df_cor` lookups per scale.
[...]

PR #183, opened 2026-06-13
perf: cache encode_pdb per-(target,atom) alignment (same output) — #180 Part A

breimanntools:
## Part A of #180 — `encode_pdb` alignment cache

The 13 `encode_pdb` encoders each re-run the **same** global `PairwiseAligner`
alignment of `target_seq` vs the chain's `atom_seq` (chain pick + every
per-feature value mapping) — ~26 `O(L^2)` alignments per entry on identical
inputs (~19 ms/entry at L=500, ~74 ms at L=1000).

This factors the alignment into an `lru_cache`d `_alignment_strings(target_seq,
atom_seq)` shared by `_identity_fraction` and `_align_atom_values_to_target`.
`aligner.align(...)[0]` is the deterministic first optimal alignment, so the
cached strings are identical → **byte-identical encoder output**; no
encoder-signature change. ~12x off the repeated-alignment overhead.

### Same-output guarantee
`tests/.../test_struct_alignment_cache_equiv.py` proves every no-external-dep
encoder is byte-identical with the cache **bypassed** vs **served from the
shared `lru_cache`**, on the bundled `P1`/`P2`/`AF_TINY` fixtures, plus a
cross-contamination guard on distinct keys. Full local suite green
(4324 passed, 11 skipped).

### Scope
- Code: `aaanalysis/data_handling_pro/_backend/struct_preproc/encode_pdb.py`
- Test: `test_struct_alignment_cache_equiv.py`
- Docs: `CHANGELOG.md` + `release_notes.rst`

Part B (pro-IO fetch `Session` reuse + opt-in concurrency) lands as a separate
PR — **#180 deliberately stays open** until both strands land (no closing
keyword here).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #180, opened 2026-06-13
perf: finish same-output optimization sweep — encode_pdb alignment cache + pro-IO fetch concurrency


breimanntools:
## Problem
The library-wide same-output performance sweep merged 9 optimizations across #169, #172, #174, #175 (≈1.5–50x; each pinned by an equivalence test, full CI matrix green). Two scoped strands remain unlanded:

1. **`encode_pdb` repeats one alignment ~26×/entry.** Its 13 encoders each re-run the *same* global `PairwiseAligner` alignment of `target_seq` vs the chain's `atom_seq` (chain pick + every per-feature value mapping) — pure redundancy of ~19 ms/entry at L=500, ~74 ms at L=1000.
2. **AlphaFold/UniProt bulk fetch is connection- and latency-bound.** A fresh TCP+TLS connection is opened per request and entries are fetched strictly sequentially.

## Goal
Land both as **same-output** (byte-identical data) optimizations, each as its own small PR with an equivalence test, under the established "benchmark-before-claiming" discipline (the audit's % estimates proved unreliable).

### A. `encode_pdb` alignment cache — code-complete, awaiting polish + PR
Local branch `feat/perf-encode-pdb-align`, commit `0a6856ac` (worktree `../aaanalysis-perf-io`).
- [x] `lru_cache` a shared `_alignment_strings(target_seq, atom_seq)` used by `_identity_fraction` + `_align_atom_values_to_target` (deterministic first alignment → byte-identical).
- [ ] Release-notes + `CHANGELOG.md` entry.
- [ ] Dedicated cache-equivalence test (encode_pdb output identical with vs without a primed cache) on the bundled `P1`/`P2`/`AF_TINY` fixtures.
- [ ] Push → PR → merge on green.

### B. pro-IO fetch (AlphaFold + UniProt)
- [ ] Refactor the HTTP seam so the suite mocks `requests.Session` (today it mocks module-level `requests.get`; ~18 tests across 3 files assert on it — this is the main cost and why the first attempt was reverted).
- [ ] Session reuse (`requests.Session`) for connection pooling — byte-identical responses.
- [ ] Opt-in concurrent fetch (`ThreadPoolExecutor`) with **input-order row reassembly** (identical DataFrame/files), **OFF by default** — parallel requests to AF-DB/UniProt risk 429 throttling that turns successes into failures (an output change); gate behind `max_workers`/`options['n_jobs']`.
[...]

PR #194, opened 2026-06-13
perf: struct_pro Batch-6 same-output wins (#186)

breimanntools:
Part of #186 (Batch 6 — whole-library same-output audit tail). **Tier: T1 — byte-identical** (ADR-0032); no regression anchor required.

Three exact same-output optimizations in `data_handling_pro/_backend/struct_preproc`, each benchmarked in isolation and pinned by a committed equivalence test:

- **`encode_disulfide`** — O(n²) SG–SG distance double-loop → vectorized pairwise distance (same formula as the merged #175 CA–CA fix). Byte-identical incl. the 2.5 Å inclusive boundary, equidistant ties, nearest-partner pick, and the NaN/NaN result for CYS-without-SG. ~16–48x (grows with cysteine count).
- **DSSP session aligner** — reuse one session-scoped `PairwiseAligner` across entries instead of 3 fresh constructions per entry (a separate copy from the #180-A encode_pdb cache). Identity fraction + alignment strings byte-identical.
- **`_plddt_per_residue` reuse** — compute once per entry, share across the 3 pLDDT encoders. Output identical. ~2.2x.

Backend-internal only — no public API/signature/docstring change (the new params are optional backend kwargs).

**Tests:** `tests/unit/data_handling_pro_tests/test_struct_batch6_equiv.py` (mirrors the #183 cache-equiv pattern; fixtures SS_BOND/AF_TINY/P1). Scoped local gate green (195 passed); `test_param_coverage` + `test_backend_import_hygiene` green.

Note: edits `release_notes.rst`, so it will conflict with the sibling Batch-6 PRs (`feat/perf-batch6-cpp`, `feat/perf-batch6-misc`) — first to merge wins, the others get synced.

### DIFF
--- a/aaanalysis/data_handling_pro/_backend/struct_preproc/align_dssp_full.py
+++ b/aaanalysis/data_handling_pro/_backend/struct_preproc/align_dssp_full.py
@@ -44,12 +44,21 @@ ChainFull = Tuple[
 
 
 def pick_best_chain_full_(
-    target_seq: str, chains: List[ChainFull]
+    target_seq: str, chains: List[ChainFull],
+    aligner: Optional[PairwiseAligner] = None,
 ) -> Optional[Tuple[ChainFull, float]]:
-    """Choose the chain whose ATOM sequence best matches ``target_seq``."""
+    """Choose the chain whose ATOM sequence best matches ``target_seq``.
+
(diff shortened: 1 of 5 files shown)


breimanntools (commented):
If output is the same, implement it

Issue #187, opened 2026-06-13
perf: committed benchmark + regression-guard suite for hot entry points


breimanntools:
## Problem
AAanalysis is performance-sensitive (CPP feature matrices, AAclust clustering, window sampling, structure encoding), yet there is **no committed way to measure performance or catch regressions**. The recent same-output optimization sweep (#169, #172, #174, #175; queue in #180/#186) had to rely on *throwaway, gitignored* `dev_scripts/perf_*_validate.py` harnesses, and the upfront audit's % estimates were wrong about half the time (FASTA writer, TreeModel CV parallelism, `encode_pae` loops all benched as no-ops despite optimistic estimates). Today, a perf regression would ship silently and future perf work is guesswork.

## Goal
A small, committed benchmark suite over the hot public entry points, runnable locally and (as a non-blocking nightly CI job) able to flag regressions — so perf work becomes data-driven and regressions are caught. Keep it **slim** (a focused set of benchmarks, not an exhaustive grid).

## Requirements
- Pick one tool and document the choice: `pytest-benchmark` (lowest overhead, lives beside tests) or `asv` (history/tracking). Recommend `pytest-benchmark` for slimness.
- Cover the genuine hot paths with deterministic, bundled fixtures (small `load_dataset`/scales + the bundled PDB fixtures):
  - `CPP.run` / `CPP.run_num`, `AAclust.fit`, `SequenceFeature.feature_matrix`,
  - `AAWindowSampler` sampling (the vectorized filters), `StructurePreprocessor.encode_pdb`,
  - `dPULearn.fit`, `TreeModel.fit`.
- Store committed baseline timings; provide a helper (mirroring `.github/scripts/check_branch_coverage.py`) that compares a run against baseline and flags a slowdown beyond a generous threshold (e.g. >1.5x) to absorb CI noise.
- Wire as a **nightly / opt-in** job (NOT the blocking matrix — wall-clock is noisy on shared CI), mirroring how the exact-value CPP regression anchor runs nightly (ADR-0015).
- Document the run command in `.claude/rules/workflow.md` and `docs/guides/agentic_engineering.md`.

## Out of scope
- Replacing the ad-hoc `dev_scripts/` harnesses (they remain for one-off exploration).
- Making perf a blocking gate on the main matrix (too noisy initially).
- Micro-optimizing — this issue is the *measurement* infrastructure only; actual optimizations live in #180/#186.

## Measurable KPIs (acceptance)
[...]

PR #196, opened 2026-06-13
perf: Batch-6 same-output wins — AAMut/sliding-window (#186)

breimanntools:
Part of #186 (Batch 6 — whole-library same-output audit tail). **Tier: T1 — byte-identical** (ADR-0032); no regression anchor required.

Two exact same-output optimizations across protein_design / seq_preproc, each pinned by a committed equivalence test:

- **`AAMut.comp_substitution_impact`** (`protein_design/_backend/aamut/aamut.py`) — replace the nested per-(from,to) `pd.DataFrame` build + `concat` with column-accumulation + a single DataFrame. `df.equals(reference)` True incl. dtypes/row-order, shape (22800,7). Public signature/return/docstring unchanged. ~19x (headline win of the batch).
- **`get_sliding_aa_window`** (`data_handling/_backend/seq_preproc/`) — inline strided window slice instead of a per-position `get_aa_window` call. Byte-identical (padding-exact form; verified it only diverges at negative starts, which the loop never produces). ~1.8x.

**Dropped from this PR:** `bic_score_` — its ~1.1–1.4x gain was marginal and the reordered float reduction was `allclose(1e-10)` only, not byte-identical. Per the same-output bar (keep only meaningful-gain **and** byte-exact), it was removed; the remaining changes are all byte-identical T1.

**Tests:** `test_aamut_substitution_impact_equiv.py`, `test_sp_get_sliding_aa_window_equiv.py`. (The known `test_to_fasta` xdist flake is unrelated and passes clean serially.)

Note: edits `release_notes.rst` → conflicts with the sibling Batch-6 PRs (#194, #195) and the merged #187 benchmark note; resolved on sync, keeping all entries.


breimanntools (commented):
Okay

PR #202, opened 2026-06-13
perf: candidate_centers_ nearest-excluded via searchsorted (O(n) memory, same output)

breimanntools:
## What
`candidate_centers_`'s vectorized band filter built an `(n_centers × n_excludes)` distance matrix — **~6.5 MB peak** on a 2000-residue / 199-exclude case (≈0 in the original scalar loop), an **O(n·m) memory blow-up that grows with input**. Replaced with `np.searchsorted` into the sorted excluded positions: the nearest excluded position is one of the two bracketing neighbours, giving **O(n_centers) memory** and O(n_centers·log n_excl) time.

## Why
A speed optimization must not trade a new O(n·m) intermediate for the gain (memory guard, ADR-0033). A per-function speed+peak-RSS audit found this was the one swept function whose vectorization materially grew peak memory.

## Result (measured)
- **Byte-identical** to the brute-force `min`-over-all-excludes (300 random cases).
- **Peak memory 6.5 MB → ~0 MB**; slightly faster than the matrix form.
- Window-sampler suite: 315 passed.

### DIFF
--- a/aaanalysis/seq_analysis/_backend/aa_window_sampler/_utils.py
+++ b/aaanalysis/seq_analysis/_backend/aa_window_sampler/_utils.py
@@ -97,11 +97,16 @@ def candidate_centers_(seq_len, window_size, exclude_positions=None,
         return centers
     if min_distance is None and max_distance is None:
         return centers
-    # Nearest-excluded L1 distance per center, vectorized. Distances are integers, so the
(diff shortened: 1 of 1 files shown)


breimanntools (commented):
If it gives the same result, push.

PR #205, opened 2026-06-13
perf(struct): share per-entry chain pick across encode_pdb encoders — byte-identical (#201)

### LINKED ISSUE #201: perf: encode_pdb shared per-entry coord/residue cache — byte-identical (#186 tail)
## Problem
`encode_pdb` runs ~13 per-residue encoders per structure; each independently re-walks the structure to collect chain residues and pick the best chain (`_collect_chain_residues` / `_pick_best_chain_records`) before encoding. The merged alignment cache (#183) covered only the *alignment* leg, not this structure-walk + chain-pick — the triage measured ~12× redundancy off it per entry.

## Goal
Extract the CA/CB coordinates + ordered residue list + best-chain pick **once per entry** and share them across all encoders — **byte-identical output**.

## Requirements
- **Tier: T1 — byte-identical** (ADR-0032): the chain pick is deterministic, so shared vs independent encoder output is identical; replace in place.
- Invasive: thread the shared cache through the encoder dispatch loop. **Preserve the per-entry error envelope** — a failure must degrade just that row (`pdb_ok=False`), not abort the whole call (cf. the #194 pLDDT lazy-compute fix).
- Committed equivalence/golden test on bundled fixtures P1/P2/AF_TINY (+ SS_BOND), pinning byte-identical encoder blocks shared-vs-independent.

## KPIs (acceptance)
- Benchmarked speedup stated and measured in isolation (expect ~12× off the redundant walk; audit %s are unreliable — confirm before claiming).
- Byte-identical encoder output across all fixtures (committed test).
- Full CI matrix green.

This is the **last open same-output item in #186**; closing it lets #186 close.
Refs: #186; #183 / #180 Part A (alignment cache); #194 (pLDDT error-envelope precedent).

breimanntools:
Closes #201 — the last open same-output (T1) tail of #186.

## Problem
`encode_pdb` runs ~13 per-residue encoders per structure. Each encoder independently re-walked the structure to collect amino-acid chain residues (`_collect_chain_residues`) and rebuilt each chain's atom sequence to pick the best-matching chain (`_pick_best_chain_records`). The #183 alignment cache covered only the O(L²) aligner, **not** this structure-walk + chain-pick — the #186 triage measured ~12× redundancy off it.

## Change (T1 — byte-identical, ADR-0032)
- **Backend** (`_backend/struct_preproc/encode_pdb.py`): add `_resolve_best_chain` (collect + best-chain pick, collapsing the former `not chains` / `chain is None` early-returns into one) and thread an optional `chain_pick` through every encoder + shared helper (`_plddt_per_residue`, `_encode_chi_sincos`, `_ca_coords_and_residues`, `_encode_contact_count`). `encode_depth` / `encode_hse` keep their ordering (resolve → `chain is None` short-circuit → external tool), so an empty structure still returns NaN rather than erroring.
- **Frontend** (`_struct_preproc.py`): the pick is computed **once per entry**, lazily inside the existing per-feature `try/except` (mirroring the #194 pLDDT precedent already in the dispatch loop). A `RuntimeError` during resolution degrades just that row (`pdb_ok=False`) with the identical `"PDB encoder '<key>' failed"` message; a non-`RuntimeError` propagates exactly as before — **byte-identical output AND error envelope**.

The pick is a deterministic function of `(structure, sequence)`; the external tools mutate `residue.xtra` / internal coords but never chain membership or resnames, so a once-computed pick equals every per-encoder recompute.

## Equivalence (committed test)
`tests/unit/data_handling_pro_tests/test_struct_chain_cache_equiv.py` pins each no-external-tool encoder byte-identical **shared-vs-independent** on **P1 / P2 / AF_TINY / SS_BOND** (36 checks), plus `_resolve_best_chain == inline collect+pick` and the empty-structure → `(None, …)` path.

## Benchmark (measured in isolation — audit %s are unreliable)
- The redundant per-encoder structure walk + chain pick is removed **N → 1 per entry** (N = requested PDB features, up to 13) — this *is* the "~12×", i.e. the **walk-redundancy factor**.
- **End-to-end** `encode_pdb` gain is **modest and feature-mix-dependent** (the per-residue value extraction, notably the two O(n²) contact-count encoders, dominates; #183 already cached the costly alignment leg). This lands as a **same-output structural-redundancy removal**, not a headline wall-clock speedup.

Per ADR-0032, T1 requires no nightly regression anchor; the committed equivalence test + existing parity suite suffice.

## Tier checklist (ADR-0032 / CONTRIBUTING)
- [x] Tier declared: **T1 — byte-identical**
- [x] Validation harness (gitignored `dev_scripts/perf_encode_pdb_chaincache_validate.py`) + committed equivalence test
- [x] One optimization per PR
- [x] Local gates green: full fast unit gate (4542 passed / 12 skipped), `data_handling_pro` + param-coverage + backend-hygiene (818 passed), docs build OK, CI lint subset clean, no new flake8-88 violations

### DIFF
--- a/aaanalysis/data_handling_pro/_backend/struct_preproc/encode_pdb.py
+++ b/aaanalysis/data_handling_pro/_backend/struct_preproc/encode_pdb.py
@@ -93,6 +93,13 @@ def _pick_best_chain_records(target_seq: str, chains):
 def _resolve_best_chain(structure, sequence: str):
     """Collect chains and pick the best-matching one **once** per entry.
 
+    "Best-matching" means the chain whose ATOM-record residue sequence has the
+    highest **identity fraction** to the target ``sequence`` — i.e. the number
+    of matched (non-gap) positions in the global alignment of the two, divided
+    by ``len(sequence)`` (see :func:`_identity_fraction`). Ties are broken by
+    chain order (the first chain reaching the maximum wins). This is the chain
+    every encoder aligns against and reads per-residue values from.
+
     ``_collect_chain_residues`` + ``_pick_best_chain_records`` are pure,
     deterministic functions of ``(structure, sequence)``, so the result can be
     computed once and shared across every encoder for the same entry instead of
--- a/aaanalysis/data_handling_pro/_backend/struct_preproc/encode_pdb.py
+++ b/aaanalysis/data_handling_pro/_backend/struct_preproc/encode_pdb.py
@@ -90,6 +90,23 @@ def _pick_best_chain_records(target_seq: str, chains):
     return chain, residues, atom_seq, best_score
 
 
+def _resolve_best_chain(structure, sequence: str):
+    """Collect chains and pick the best-matching one **once** per entry.
+
+    ``_collect_chain_residues`` + ``_pick_best_chain_records`` are pure,
+    deterministic functions of ``(structure, sequence)``, so the result can be
+    computed once and shared across every encoder for the same entry instead of
+    re-walking the structure (and rebuilding each chain's atom sequence) ~13x.
+    Returns ``(chain, residues, atom_seq, identity)``; ``chain`` is ``None``
+    when the structure has no amino-acid chains — the encoders map that to an
+    all-NaN block with identity 0.0, collapsing the former ``not chains`` and
+    ``chain is None`` early-returns into one check."""
+    chains = _collect_chain_residues(structure)
+    if not chains:
+        return None, None, None, 0.0
+    return _pick_best_chain_records(sequence, chains)
+
+
 def _align_atom_values_to_target(target_seq: str,
                                  atom_seq: str,
                                  atom_values: List[float]) -> List[float]:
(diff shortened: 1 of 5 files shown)


aaanalysis/data_handling_pro/_backend/struct_preproc/encode_pdb.py
```diff
@@ -90,6 +90,23 @@ def _pick_best_chain_records(target_seq: str, chains):
     return chain, residues, atom_seq, best_score
 
 
+def _resolve_best_chain(structure, sequence: str):
+    """Collect chains and pick the best-matching one **once** per entry.
```
breimanntools, line 94: What does best matching one here mean? Define please

breimanntools (commented):
All good beside this one clarification.

aaanalysis/data_handling_pro/_backend/struct_preproc/encode_pdb.py
breimanntools: Defined in the docstring (7dcf627): "best-matching" = the chain whose ATOM-record residue sequence has the highest **identity fraction** to the target `sequence` — matched non-gap positions in the global alignment / `len(sequence)` (see `_identity_fraction`); ties broken by chain order (first chain reaching the max wins).

Issue #65, opened 2026-04-28
Biopython-backed MSA generation pipeline (homolog search + alignment)


breimanntools:
## Problem

The current `get_msa` function is misnamed: it downloads a single FASTA sequence from UniProt for a given accession. This is not a multiple sequence alignment — it's one sequence. To compute meaningful conservation scores, AAanalysis needs an actual MSA built from homologs of the input sequence.

## Goal

Provide a Biopython-backed pipeline that takes a UniProt ID (or raw sequence) and produces a true MSA, with pluggable backends for both homolog retrieval and alignment.

## Tasks

**Homolog retrieval (pick at least one, ideally two, as backends):**
- BLAST (NCBI remote or local)
- HMMER / jackhmmer
- MMseqs2 (fast, recommended for large-scale)
- ColabFold / MMseqs2 API (cloud-hosted, no local install)

**Alignment (pick at least one, ideally two):**
- Clustal Omega
- MUSCLE
- MAFFT
- T-Coffee (optional, slower)

**Pipeline scaffolding:**
- `aa.get_msa(uniprot_id=..., search_tool=..., align_tool=..., n_seqs=..., e_value=...)`
- Use the existing `check_is_tool` helper to verify external dependencies are on `PATH`
- Cache MSAs locally to avoid recomputation (configurable cache dir)
- Filter alignments (e.g., `--max-seqs`, identity range, gappy-column removal) before returning
- Support both file output (`output_file=...`) and in-memory `Bio.Align.MultipleSeqAlignment` return
- Document which backends are required vs. optional, and how to install them (or fall back to cloud APIs if nothing is local)

**Tests & docs:**
- Unit tests with mocked backends
- Integration test against the cloud MMseqs2 API for one small example
- Tutorial showing the full pipeline from UniProt ID → MSA → conservation scores

## How this improves AAanalysis

- Makes the conservation feature actually correct (an MSA from one sequence is meaningless)
- Gives users flexibility: fast cloud backend for casual use, local tools for batch / production
- Brings AAanalysis in line with how other interpretable-protein-prediction tools handle evolutionary information (e.g., EVcouplings, ESM-MSA)

## Acceptance criteria

- `aa.get_msa("P12345")` returns a valid MSA with ≥10 homologs by default
- At least one search backend and one alignment backend work end-to-end without local installs (cloud fallback)
- Backends are documented with install instructions and a "which should I use?" decision matrix
[...]

PR #224, opened 2026-06-14
feat(security): subprocess/network/file hardening audit (#88)

### LINKED ISSUE #88: Code-security hardening audit (subprocess / network / file parsing)
## Problem

AAanalysis has a real code-security surface that no issue tracks. It shells out
to external tools and fetches from the network, and parses untrusted files:

- **Subprocess:** FIMO (`seq_analysis_pro/_scan_motif.py`), the shared runner
  `seq_analysis_pro/_backend/_utils.py::run_command` (`subprocess.Popen`),
  CD-HIT (`_backend/cd_hit.py`), MMseqs2 (`_backend/mmseq2.py`), ChainSaw
  (`data_handling_pro/_backend/struct_preproc/_chainsaw.py`), plus the planned
  MSA backends (#65: BLAST / MMseqs2 / Clustal / MUSCLE).
- **Network fetch:** UniProt
  (`data_handling_pro/_backend/annot_preproc/_uniprot.py:81`), AlphaFold DB
  (`data_handling_pro/_backend/struct_preproc/_alphafold.py:73`), and a
  conservation fetch (`seq_analysis_pro/_comp_seq_cons.py:66` — note: this
  `requests.get(url)` has **no `timeout`**, unlike the other two).
- **File parsing:** FASTA / PDB / mmCIF / xlsx / tsv treated as trusted.

Good news from a first read: every subprocess call already uses an **argument
list** (no `shell=True`, no f-string command). So the audit's job is to
*confirm and pin* that, close the specific gaps (e.g. missing HTTP timeout,
unvalidated accessions/URLs, unbounded response sizes, path traversal on
downloaded/parsed files), and document the trust boundary — not a rewrite.

## Goal

Audit and harden the subprocess / network / file-parsing paths so user-supplied
input can never reach a shell, downloads/inputs are validated and bounded, and
the trust boundary is documented — **without adding a `SECURITY.md`** (hard-rule
reject per `.claude/rules/sharp-edges.md`); CodeQL continues in CI.

## Requirements

- [ ] Subprocess: confirm **0** call sites use `shell=True` or string-built
      commands; verify FIMO/CD-HIT/MMseqs2/ChainSaw args are arg-lists with
      numeric/option values cast explicitly (`str(int(...))`/`str(float(...))`),
      and that temp files come from `tempfile` (already the pattern in `_utils`).
- [ ] Network: give **every** `requests.get` an explicit `timeout=` (fix
      `_comp_seq_cons.py`); validate UniProt/AlphaFold accessions against an
      expected pattern before interpolating into the URL; bound/stream response
      size and fail closed on unexpected content type/status.
- [ ] File parsing: treat downloaded + user files as untrusted — guard output
      paths against traversal (resolve + confine to the target dir) and raise
      `ValueError`/`RuntimeError` on malformed FASTA/PDB/mmCIF rather than
      silently mis-parsing.
- [ ] Add a short docstring/doc note on the network/subprocess **trust
      boundary** (which inputs are sent to external tools / remote servers).
- [ ] Produce a checklist (in the PR / a DEV comment) of every audited call site
      with its disposition (OK / hardened).

## KPIs / Acceptance criteria

- [ ] `grep -rn "shell=True" aaanalysis/` returns **0**, and an audit checklist
      covers **all 6 call-site groups** (FIMO, CD-HIT, MMseqs2, ChainSaw,
      UniProt, AlphaFold + conservation HTTP), each marked OK/hardened.
- [ ] **100%** of `requests.get` calls in `aaanalysis/` pass an explicit
      `timeout` (asserted by a grep-based meta-test or unit test).
- [ ] An accession/URL-validation test rejects a crafted bad accession (e.g.
      `"../../etc"` or one with URL metacharacters) with `ValueError` before any
      network call; a path-traversal test rejects an out-of-dir output path.
- [ ] **No `SECURITY.md`** is added (explicitly verified).

## Scope / non-goals

- Pro/external-tool wrappers + data-handling-pro fetchers (where the surface
  lives). **Out of scope:** a `SECURITY.md` vulnerability-reporting policy
  (hard-rule reject); changing CodeQL config.

## Dependencies

- From the gap analysis. Coordinates with #65 (MSA backends inherit the same
  subprocess rules). Most touched code is `*_pro` (gated behind the `pro` extra).

## Standards checklist

- [x] backend trusts frontend, but **external/remote input is untrusted** —
      validate at the boundary
- [ ] CONFIRM-FIRST: none (no `pyproject`/`__init__`/workflow change expected)
- [x] no `print()`; bare `ValueError`/`RuntimeError`; `requests` is a `pro` dep
- [x] tests: validation/traversal/timeout

breimanntools:
Code-security hardening audit of subprocess / network / file-parsing call-sites (issue #88), run as an overnight session lane. Audit + small, behavior-preserving fixes only — **not** a SECURITY.md (sharp-edges reject) and **no** custom exception base.

## Findings
All subprocess call-sites (FIMO, CD-HIT, MMseqs2, ChainSaw) already use list-args (no `shell=True`); network fetches route through a shared transport with timeouts; parsers are bounded/trusted. `grep shell=True` → 0.

## Fixes applied
- **Added `timeout=30.0`** to `get_msa` in `seq_analysis_pro/_comp_seq_cons.py` (was a bare `requests.get`) — the issue-flagged gap
- New AST meta-test `tests/unit/api_tests/test_network_timeout_hygiene.py` enforcing every `requests.get`/session `.get`/`http_get_` carries an explicit `timeout=` (with a self-check)
- Audit doc `docs/guides/security_audit_88.md`

74 tests green. No CONFIRM-FIRST surface touched.

## Left as findings (behavior-changing → your call, see audit doc)
- F1 accession-pattern validation · F2 `run_command` broad-except cleanup · F3 `_comp_seq_cons.py` orphan wire-up-or-remove · F4 response-size bounding (defer to #65)

Closes #88

### DIFF
--- a/aaanalysis/seq_analysis_pro/_comp_seq_cons.py
+++ b/aaanalysis/seq_analysis_pro/_comp_seq_cons.py
@@ -41,9 +41,13 @@ def normalize_scores(scores: list) -> np.ndarray:
 
 
 # II Main Functions
-def get_msa(uniprot_id: str, output_file: str = "msa.fasta") -> str:
+def get_msa(uniprot_id: str, output_file: str = "msa.fasta",
+            timeout: float = 30.0) -> str:
     """
     Retrieve multiple sequence alignment (MSA) from UniProt.
+
+    Trust boundary: ``uniprot_id`` is interpolated into a remote URL and the
+    response body is written to disk. Treat the fetched content as untrusted.
 
     Parameters
     ----------
@@ -51,6 +55,10 @@ def get_msa(uniprot_id: str, output_file: str = "msa.fasta") -> str:
         UniProt ID of the target protein.
     output_file : str, default="msa.fasta"
         File path to save the retrieved MSA.
+    timeout : float, default=30.0
+        Seconds to wait for the UniProt response before failing (passed to
+        ``requests.get``). A finite default is enforced so the call can never
+        hang forever on an unresponsive server.
 
     Returns
     -------
@@ -63,7 +71,7 @@ def get_msa(uniprot_id: str, output_file: str = "msa.fasta") -> str:
         If the MSA retrieval fails.
     """
     url = f"https://www.uniprot.org/uniprot/{uniprot_id}.fasta"
-    response = requests.get(url)
+    response = requests.get(url, timeout=timeout)
 
     if response.status_code == 200:
         with open(output_file, "w") as f:
--- a/docs/guides/security_audit_88.md
+++ b/docs/guides/security_audit_88.md
@@ -1,0 +1,112 @@
+# Code-security hardening audit (issue #88)
+
+This document records the audit of AAanalysis's code-security surface —
+subprocess invocations, network fetches, and untrusted file parsing — and the
+small, behaviour-preserving fixes applied. It is **not** a vulnerability-
+reporting policy: per `.claude/rules/sharp-edges.md`, no `SECURITY.md` is added
+and none is intended.
+
+Most of the surface lives in the `*_pro` subpackages (gated behind the `pro`
+extra: biopython, requests, the MEME suite for FIMO, external CLI binaries).
+
+## Trust boundary (summary)
+
+- **Backend trusts the frontend** for argument validation (house rule). But
+  **external / remote input is untrusted**: anything fetched over the network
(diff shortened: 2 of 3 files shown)


breimanntools (commented):
Okay

Issue #157, opened 2026-06-13
dPULearn.fit: accept 1/0 labels directly + n_neg alias (drop manual PU-encoding)


breimanntools:
## Problem
`dPULearn.fit` requires PU-encoded labels — `1` (positive) / `2` (unlabeled) — and rejects anything else (`check_labels(..., vals_required=[1, 2], allow_other_vals=False)`). But the bundled datasets and the rest of the API use **1/0** (positive/negative): `load_dataset` documents *"0 for negatives, 1 for positives"*. So every user must hand-encode before fitting:
```python
labels_pu = [1 if x == 1 else 2 for x in labels]
n_pos = sum(np.array(labels_pu) == 1)
dpul.fit(X, labels=labels_pu, n_unl_to_neg=n_pos // 2)
```
Passing raw 1/0 labels **raises**, and the count arg `n_unl_to_neg` ("number of unlabeled to reclassify as negative") is routinely conflated with "number of negatives". This is an avoidable footgun and forces a 3-line dance in every example (incl. the cheat sheet).

## Goal
Let `dPULearn.fit` accept standard 1/0 labels via an explicit, opt-in mapping (treat 0 as unlabeled for mining), and expose `n_neg` as the primary name for `n_unl_to_neg` — without changing behavior for existing 1/2 callers.

## Requirements
- [ ] Add `label_unl: int = 2` (and/or `label_pos: int = 1`) to `dPULearn.fit` (`aaanalysis/pu_learning/_dpulearn.py`); when `label_unl=0`, accept 1/0 input and normalize to the internal 1/2 contract before the existing logic.
- [ ] Add `n_neg` as the primary count param; keep `n_unl_to_neg` as a deprecated alias emitting one `DeprecationWarning`, mapped internally.
- [ ] Frontend validation block normalizes labels/aliases once; backend sees only the canonical 1/2 + count.
- [ ] numpydoc updated (named `Returns`/attributes, per-method `Examples` include) to show the 1/0 entry path.
- [ ] Update the dPULearn example notebook + cheat-sheet recipe to the simplified call.

## KPIs / Acceptance criteria
- [ ] `dPULearn().fit(X, labels=<1/0 array>, label_unl=0, n_neg=10)` yields `labels_`/`df_pu_` byte-identical to today's `labels=<1/2>, n_unl_to_neg=10` path (regression test, fixed `random_state`).
[...]

PR #225, opened 2026-06-14
refactor(dpulearn): group fit params + allow negative label markers

breimanntools:
Polish on the unreleased v1.1.0 `dPULearn.fit` label-handling feature (#157). Two changes, no behavior change to existing keyword calls.

## 1. Parameter reorder
`fit` parameters are regrouped so they read in logical order — encoding → count → algorithm:

```python
fit(self, X, labels=None,
    label_pos=1, label_unl=2, label_neg=None,   # encoding
    n_neg=None, n_unl_to_neg=None,               # count — exactly one
    metric=None, n_components=0.80)              # algorithm
```

Docstring `Parameters` reordered to match. Signature-only; every call site (tests, examples) uses keywords, and the feature is unreleased — so no deprecation.

## 2. Allow negative label markers
`check_match_labels_markers` now validates marker **type** via `ut.check_number_val` (any int) instead of `check_number_range` (which forced `min_val=0`). So `label_unl=-1` (etc.) works. Markers must still be distinct and match the labels; **output stays canonical** (`0` = negative, `1` = positive, `2` = unlabeled). New negative-marker test added.

## Verified
Full fast suite **4598 passed / 14 skipped**; `test_param_coverage` ✓; docstring checker 0 defects / 0 drift. No CHANGELOG change (polish on the unreleased feature; existing entry already says "arbitrary positive/unlabeled/negative encoding").

### DIFF
--- a/aaanalysis/pu_learning/_dpulearn.py
+++ b/aaanalysis/pu_learning/_dpulearn.py
@@ -33,9 +33,10 @@ def check_n_to_identify(labels=None, n_to_identify=None, label_unl=None):
 
 def check_match_labels_markers(label_pos=None, label_unl=None, label_neg=None):
     """Validate the positive/unlabeled/negative marker values used to encode the input labels."""
-    ut.check_number_range(name="label_pos", val=label_pos, min_val=0, just_int=True)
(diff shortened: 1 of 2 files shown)


breimanntools (commented):
Good to go

PR #233, opened 2026-06-23
fix(errors): list available options in data-backed-lookup error messages

breimanntools:
## What

Enriches the user-facing errors raised when a **name is looked up against a DataFrame the caller passed in** and not found, so an agent (or human) that gets a name wrong sees actionable candidates instead of a dead end.

- New `ut.preview_options` helper — renders the first 5 valid values + total count, so a large user DataFrame never dumps thousands of ids.
- Three entry-lookup errors now append `Available entries/names: [...]`:
  - `SeqMut` mutation entry vs `df_seq`
  - `SequenceFeature` `df_parts` sample vs `df_seq`
  - CPP plot `names_to_show` vs `df_seq`
- Two wording-drift sites aligned to the house `"should be one of"` phrasing:
  - `TreeModel.check_metrics`
  - `plot_gco` invalid-option error

## Why

The bundled-data lookups (`load_dataset` / `load_scales` / `load_features`) already list their valid options; the residual gap was lookups against user-provided DataFrames, which previously said only "X is not in df" with no next step.

## Scope / safety

- **Behavior-preserving** — message text only; no control-flow or signature changes.
- Touches only library `_*.py`, the `ut` barrel import line, and tests. No CONFIRM-FIRST surfaces.

## Tests

- New `tests/unit/utils_tests/test_preview_options.py` (8 tests: short/exact/truncated/empty/Series/dict-keys/custom-n + an end-to-end enriched-message smoke check).
- Updated the one `plot_gco` test that pinned the old "not valid" wording.
- Affected suites run locally green: `sequence_feature`, `tree_model`, `cpp_plot`, `plotting`, `protein_design`, `utils_tests`, `api_tests`, integration `test_seq_pipeline`.


breimanntools (commented):
All good

PR #235, opened 2026-06-23
test(smoke): add a sub-minute critical-path smoke gate

breimanntools:
## What
Adds a `smoke` pytest marker and one tiny, seeded end-to-end check of the spine most changes touch:
`load → parts → CPP.run → feature_matrix → TreeModel.fit` (+ `predict_proba`).

`pytest -m smoke -c tests/pytest.ini` finishes in **~7s** locally and catches a gross break (import error, empty/malformed `df_feat`, model that won't fit) before the full suite.

- Register the `smoke` marker in `tests/pytest.ini`.
- `tests/smoke/test_smoke_pipeline.py` reuses `tests/_pipeline` (call pattern defined once); asserts the `df_feat` canonical contract columns, design-matrix shape, a fitted model with feature importances, and `predict_proba` shape.
- Documented in `workflow.md`.

## Scope / safety
- **Non-blocking** — deliberately NOT a separate CI job (the full matrix already covers these paths); a developer-convenience gate only. No `.github/workflows/*` or other CONFIRM-FIRST surface touched.

## Verification
- `pytest -m smoke -c tests/pytest.ini`: 2 passed, 4729 deselected in 4.8s (7.2s wall incl. startup).
- Marker registers cleanly (no unknown-marker warning).

## Note for reviewer
Second of a **stacked set** (5 → **7** → 8 → 9). **Base is `feat/df-feat-contract` (PR #234)**, so the diff here is smoke-only; GitHub will retarget this to `master` once #234 merges.

### DIFF
--- a/.claude/rules/workflow.md
+++ b/.claude/rules/workflow.md
@@ -11,6 +11,11 @@ uv pip install -e ".[dev]"
 # run all tests
 pytest tests -x -vv --durations=20
 
+# sub-minute smoke gate — the critical CPP path end to end
+# (load -> CPP.run -> TreeModel.fit). Fast local sanity before the full suite;
+# finishes in a few seconds. Non-blocking (no dedicated CI job).
+pytest -m smoke -c tests/pytest.ini
+
 # run a single file or method
 pytest tests/unit/aa_window_sampler_tests/test_aa_window_sampler_synthetic.py -x -vv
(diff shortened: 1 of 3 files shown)


breimanntools (commented):
This is good as minimal smoke test. Push to master

PR #236, opened 2026-06-23
refactor(utils): extract constants into _constants.py behind the import barrel

breimanntools:
## What
The first and largest slice of the long-deferred `utils.py` split, done with the **barrel pattern** so it is mechanically safe.

Moves the inline constant block (~470 lines: folder/file names, part/split/scale vocabularies, column names, `df_feat`/`df_eval` columns, labels, colours, option strings) out of `utils.py` into a new sibling **`aaanalysis/_constants.py`**, re-exported via `from ._constants import *`.

- `utils.py` stays the single public access point (`import aaanalysis.utils as ut`) and drops from **1073 → 610 lines**.
- **None of the ~2400 `ut.X` references change.**
- `_constants.py` imports only the stdlib + numpy — never `aaanalysis.utils` — so it loads first with no circular dependency.
- The validators are already split under `_utils/check_*.py` (and re-exported), so this is the **constants slice**; the remaining inline functions (plotting glue, data-io) can follow in a later slice.

## Why now
`utils.py` was the documented "god module." The barrel keeps the access point stable while shrinking the file, improving navigability for both humans and agents — without the blast radius of rewriting call sites.

## Safety / verification
- **Name-set invariance**: the public surface of `ut` is byte-identical before/after — **353 names, none added or dropped** (verified against a pre-split snapshot).
- **Full suite green**: `pytest tests -m "not network and not regression" -n auto` → **4713 passed, 9 skipped** (every `ut.X` reference still resolves). One transient failure on the first run was the known `read_fasta` xdist file-race flake (passes in isolation and on re-run), unrelated to this change.
- **New guard**: `tests/unit/api_tests/test_utils_barrel.py` pins the circular-import guard (no `aaanalysis` import in `_constants`) and that every public constant is re-exported identically by `ut`.
- `sharp-edges.md` updated to describe `utils.py` as a thinning barrel.
- No CONFIRM-FIRST surface touched (no `__init__.py`/`__all__`/`pyproject`/workflow change). The diff is a clean block-move: `utils.py` −471/+4, plus the new file.

## Note for reviewer
Third of a **stacked set** (5 → 7 → **8** → 9). **Base is `test/smoke-path` (PR #235)**; GitHub will retarget to `master` as the lower PRs merge. This is the highest-blast-radius change of the set — the invariance test + full-suite green are the safety net.


.claude/rules/sharp-edges.md
```diff
@@ -3,10 +3,17 @@
 Not path-scoped. Referenced from the root CLAUDE.md. Accepted architectural
 limitations — do not "fix" them opportunistically.
 
-- **`utils.py` is a god module (~1500 lines).** Add new constants/validators
-  to the appropriate topical block (search for the nearest `# `-style
-  section header). Splitting into `_constants.py`, `_checks/`,
-  `_plotting_glue.py`, `_data_io.py` is a v2 refactor.
+- **`utils.py` is a thinning import barrel.** The bulk of the constants now live
```
breimanntools, line 6: Comments of the development (e.g., "the constants now live...") of the package are for ADR and not the actual code, or? Please adjust the docstring here

breimanntools (commented):
Thanks. Looks very good

PR #237, opened 2026-06-23
refactor(types): import matplotlib Axes/Figure/Line2D from canonical modules (pyright slice)

breimanntools:
## What
First **safe slice** of the advisory-pyright burn-down (the program's non-blocking follow-up).

The plot classes annotate types as `plt.Axes` / `plt.Figure` / `plt.Line2D`, which pyright flags under `reportPrivateImportUsage` ("not exported from `matplotlib.pyplot`"). This imports each type from its canonical public module (`matplotlib.axes` / `matplotlib.figure` / `matplotlib.lines`) and uses the bare name in annotations and docstrings, across 9 files.

## Why this slice
- **Runtime-identical** — `plt.Axes is matplotlib.axes.Axes`; the change is import/annotation only, zero behavior risk. Ideal for an unattended run.
- **Clears a whole category** — `reportPrivateImportUsage` goes from 47 → **0**; total pyright errors **1250 → 1203**.

## What's deliberately left
The dominant categories — `reportArgumentType` (564) and `reportOptional*` (~426) — come from the pervasive `param: Optional[T] = None` sklearn pattern where the runtime Validate block guarantees non-None but pyright can't see it. "Fixing" those means per-site narrowing (`assert`s) or signature changes — **behavior-adjacent judgment calls** not safe to bulk-apply unattended. `reportReturnType` (33) is a mix of real-Optionals and false positives, also per-site. These are left for an interactive follow-up. pyright is **advisory / non-blocking**, so a partial reduction is correct and safe.

## Verification
- pyright: errors 1250 → 1203; `reportPrivateImportUsage` 0.
- Plot suites green: **1112 passed** (cpp_plot, aaclust_plot, dpulearn_plot, plotting, protein_design, seq_analysis).
- Package imports cleanly (new bare-name annotations resolve at runtime).
- No CONFIRM-FIRST surface touched.

## Note for reviewer
Fourth and last of the **stacked set** (5 → 7 → 8 → **9**). **Base is `refactor/utils-barrel-split` (PR #236)**; GitHub retargets to `master` as the lower PRs merge.

### DIFF
--- a/aaanalysis/_utils/plotting.py
+++ b/aaanalysis/_utils/plotting.py
@@ -4,6 +4,7 @@ This is a script for the backend of the plotting module functions used by other 
 import seaborn as sns
 import matplotlib as mpl
 from matplotlib import pyplot as plt
+from matplotlib.lines import Line2D
 import matplotlib.lines as mlines
 import warnings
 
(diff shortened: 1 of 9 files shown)


breimanntools (commented):
This looks cleaer now. But I do not know whether this is only cosmetic. Simply push to master

PR #248, opened 2026-06-24
fix(tooling): session-title hook — immediate + worktree-aware from master

breimanntools:
## Problem

The terminal-title Stop hook keyed the topic off the session `cwd`, so a session launched from the **main checkout** (on `master`) — the common case, since sessions `cd` into per-task worktrees per-command — always resolved to the bare repo name (`aaanalysis`) and never updated. A warm-up gate (~1% context) also delayed any title.

## Fix

- **Immediate:** drop the readiness gate — the title is maintained from the first turn.
- **Worktree-aware from master:** on a master/main/detached checkout, attribute the session to the feature worktree **whose path appears most recently in this session's own transcript** (the `cd`/edit targets a Stop hook can't see from a master `cwd`). This correctly distinguishes parallel sessions each driving a different worktree — a plain "most-recent-commit" guess mislabels one session with another's branch (verified: it picked a sibling session's branch). Falls back to the repo dir name only when the transcript names no feature worktree.

Hook output stays `/dev/tty`-only and always exits 0, so it can never block a turn.

## Deployment note

The hook runs the file from the **main checkout**, so this goes live for new sessions once merged to `master`.

## Verification

`py_compile` clean; ran directly across three scenarios (master cwd + transcript → correct worktree; master cwd + no transcript → repo name; feature-branch cwd → own branch + PR#).

### DIFF
--- a/.github/scripts/set_session_title.py
+++ b/.github/scripts/set_session_title.py
@@ -1,25 +1,32 @@
 #!/usr/bin/env python3
 """Give every Claude Code session a descriptive terminal tab title, so parallel
-sessions running in different git worktrees are tellable apart at a glance.
+sessions (each driving its own git worktree) are tellable apart at a glance.
 
-Wired as a Stop hook (fires when Claude finishes a turn). Once the session has
-used ~1% of the context window it starts maintaining the tab title; it only
-re-writes the title when the computed label actually changes, so it is silent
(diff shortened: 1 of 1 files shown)


breimanntools:
Automated gates green; hook fix verified across master-cwd / worktree / feature-branch scenarios. Merging per maintainer go-ahead.

Issue #288, opened 2026-06-28
CPPStructurePlot: integrated-prediction linked explorer with selectable outputs (widget / self-contained HTML / static)


breimanntools:
## Problem
The deployed cleavage app lets a user explore per-site CPP-SHAP impact with the **feature map
and 3D structure linked** (select/hover a feature-map column → the residue lights up in the
structure, and vice-versa), re-predicting per site, and export a figure. The package now has the
static pieces (`map_structure`, `plot_combined`, `interactive` site-slider) and app-exact
colour/size/fade/zoom, but **not** the feature-map↔structure linking, the integrated
prediction, or selectable output types.

## Goal
Given a provided **`df_feat` + ML model + data** (`df_seq` + `labels`), one entry point does the
whole loop — per-site CPP + ShapModel re-prediction, render, and **feature-map↔structure
linking** — with a user-selectable **output type** so the same call serves iterative
exploration and a publication figure.

### Stage A — integrated predictor
- [ ] A built-in predictor: from `df_seq` + `labels` + a fitted `model` (+ window geometry),
  compute per-site `df_feat` (CPP features + `ShapModel.add_feat_impact`) and the predicted
  probability. Keep the custom `predictor=` callback as an escape hatch.
### Stage B — notebook interactive link
- [ ] In `interactive()`: select a feature-map position (click/slider) → highlight that residue
  in the py3Dmol structure (marker/bold stick) and mark the feature-map column; site slider +
  colour/focus toggles as today. Debounced.
### Stage C — self-contained linked HTML (app-like)
- [ ] An export that writes a standalone `.html`: feature map as an interactive SVG/canvas +
  3Dmol.js, with hover/click linking column↔residue (ports the app’s linking). Shareable;
  paper-figure source.
### Stage D — static side-by-side (paper)
- [ ] `plot_combined` already renders the structure + feature map as one figure; ensure the
  chosen site/zoom is easy to capture (PNG/PDF) and visually consistent with A–C.
### Output selection
- [ ] One surface to pick the output: `output="widget"|"html"|"static"` (or peer methods) over
  the same inputs and the same per-residue impact, so all three agree.

## KPIs / Acceptance
- [ ] Linking: selecting feature-map position P → the structure highlights the residue
  `site_to_start(p1)`-mapped from P (asserted for ≥1 case per output type).
- [ ] Integrated predictor: `df_seq`+`labels`+`model` → a per-site `df_feat` equal to a manual
[...]

PR #293, opened 2026-06-29
docs(feature_engineering_pro): mark CPPStructurePlot as experimental

breimanntools:
## Summary

Adds an `.. warning:: Experimental` admonition to the `CPPStructurePlot` class docstring. The
class and its methods (`map_structure` / `plot_combined` / `plot_linked` / `interactive` /
`explore`) are under active development; their API may change between minor releases without the
usual deprecation cycle, so this pro surface is **not** yet covered by the package's strict semver
guarantee.

Docstring-only change (no code/behaviour change). Relates to the still-open #288 (Stages C/D
pending) — the issue stays open.


breimanntools:
✅ Docstring-only experimental marker on CPPStructurePlot. Docstring checker clean. Merging on green. #288 stays open.

Issue #310, opened 2026-06-30
feat: aa.plot_eval_heatmap — house-preset evaluation heatmap (consolidate duplicated seaborn blocks)


breimanntools:
Part of #305.

## Problem

The same annotated evaluation heatmap is hand-built from seaborn every time a
sweep result is shown. In the γ-secretase notebook the **identical** block
appears in cells 12 and 25 (differing only in axis labels), and the original
project repeats it a further **3×** (`_plot_evaluation` in `plot_cpp_pred.py`,
`ml_benchmarking.py`, `pred_summary.py`):

```python
ax = sns.heatmap(df_eval, vmin=50, vmax=100, cmap="viridis", annot=True, fmt=".0f",
                 linewidth=0.1, cbar_kws=dict(label="Balanced accuracy [%]"))
ax.tick_params(left=False, bottom=False); plt.yticks(rotation=0); plt.xticks(rotation=0)
```

Five copies of the same vmin/vmax/cbar/tick fiddling.

## Goal

Provide `aa.plot_eval_heatmap(df_eval, ...)` with the house preset
(`vmin=50, vmax=100, cmap="viridis", annot, fmt=".0f", cbar label`), collapsing
all copies to one call — and reconcile it with the existing `aap.plot_eval` so
there is a single static-grid heatmap entry point, not two.

## Requirements
- [ ] `aa.plot_eval_heatmap(df_eval, xlabel=None, ylabel=None, vmin=50, vmax=100, cbar_label=..., ax=None)`
      in the plotting layer, returning the `Axes` (no `plt.show()` in library).
- [ ] House preset defaults; all overridable.
- [ ] Decide: standalone helper vs a mode of `aap.plot_eval` — document the
      relationship so they don't diverge.
- [ ] numpydoc + `Examples` include; follows `plotting.md` (no `plt.show`/
      `tight_layout` in library code).

## KPIs / Acceptance criteria
- [ ] Re-authoring notebook cells 12/25 on top of it removes the duplicated
      seaborn blocks and renders an equivalent figure (visual + shape check).
- [ ] ≥1 unit test (returns an `Axes`; respects `vmin`/`vmax`/labels; bad input
      → `ValueError`).

## Scope / non-goals
- A static heatmap; the adaptive eval-grid (`aap.plot_eval`, PR #266) stays as
  is — this is its simple sibling.

## Dependencies
- Part of epic #305 · relates `aap.plot_eval`.

## Standards checklist
[...]

Issue #309, opened 2026-06-30
feat: aa.eval_features — model + CV + metric feature-set scorer (incl. PU mask-CV)


breimanntools:
Part of #305.

## Problem

Benchmarking a feature set is the recurring spine of the γ-secretase analysis,
yet the library ships no scorer — so the notebook carries a whole sklearn import
block and a hand-rolled evaluator, called **5×** (cells 12, 24, 27):

```python
def balanced_acc(X, y):
    y_pred = cross_val_predict(SVC(kernel="linear"), X, y, cv=LeaveOneOut(), n_jobs=-1)
    return balanced_accuracy_score(y, y_pred) * 100
```

The original project additionally needs a PU-aware variant that **masks known
positives** during CV (`cv_leave_one_out_masked` in `ml_benchmarking.py`). Every
benchmark, sweep cell, and baseline funnels through this missing primitive.

## Goal

Ship `aa.eval_features` — a model + CV + metric feature-set scorer whose default
reproduces the notebook's linear-SVM / LeaveOneOut / balanced-accuracy recipe —
adding **no new required dependency** (sklearn is already a dep) and honoring the
`random_state` contract.

## Requirements
- [ ] `aa.eval_features(X, labels, model=None, cv=None, metric="balanced_accuracy", random_state=None)`
      returning the score (percentage by default), defaulting to linear-SVM +
      `LeaveOneOut` + balanced accuracy.
- [ ] Accept an sklearn estimator and a CV splitter (or int) so users can swap in
      RF / repeated-CV without leaving the API.
- [ ] PU **mask-known-positives** CV option (port `cv_leave_one_out_masked`).
- [ ] Deterministic given `random_state`; numpydoc + `Examples` include.
- [ ] Re-export in `aaanalysis/__init__.py` (CONFIRM-FIRST).

## KPIs / Acceptance criteria
- [ ] Reproduces the notebook's current balanced-accuracy numbers within
      numerical tolerance on the canonical fixture.
- [ ] Same `random_state` → identical score across runs (reproducibility test).
- [ ] ≥1 unit test per parameter (model / cv / metric), positive + negative.

## Scope / non-goals
- A scorer, not a model-selection sweep (that is `aap.find_features` / #305 item
  for the eval heatmap). Coordinate with #91 so the eval/comparison API is one
  surface, not two.

## Dependencies
- Part of epic #305 · relates #91 (model evaluation & comparison).

## Standards checklist
[...]

Issue #305, opened 2026-06-30
epic: use-case convenience methods — collapse γ-secretase notebook boilerplate


breimanntools:
## Problem

The γ-secretase Use Case (`use_cases/use_case1_gamma_secretase.ipynb`) is our
flagship end-to-end story, yet it repeatedly hand-rolls the *same* glue that the
library should own. An audit of the notebook (and the original research project
it is based on, `G-Sec_Substrate_Prediction`) found the identical
**"featurize `df_seq` → score with a CV evaluator → drop into an `np.zeros`
grid → seaborn heatmap"** loop re-implemented in cells 12 and 24, a sibling
manual grouped barplot in cell 28, and per-protein SHAP plumbing in cells 30/32.

Concretely, the notebook carries a full sklearn import block
(`SVC` + `LeaveOneOut` + `balanced_accuracy_score`) only because the library
ships no benchmark scorer; writes `sf.feature_matrix(features=…,
df_parts=sf.get_df_parts(df_seq=…, list_parts=…), df_scales=…)` in **5 cells**;
duplicates one seaborn heatmap block **3×** (and the original project a further
3×); and string-templates `col_imp=f"feat_impact_{name}"` to couple every SHAP
plot to `add_feat_impact`'s naming by hand. Every line of that is boilerplate a
user must rediscover, and it makes the canonical analysis look harder than it is.

## Goal

Ship a small set of convenience methods that collapse the recurring notebook
boilerplate into one-call APIs — **without changing any existing default output**
(byte-identical regression) and **without adding new required dependencies**
(SHAP pieces stay in the `pro` extra). Success = the γ-secretase notebook is
re-authored to use them and gets materially shorter.

## Requirements

This is a **tracking epic**: each item below is an independently-mergeable
deliverable with its own KPI and should land as its own PR (and may be split
into its own child issue). Source cell/file references are given so any lane is
self-contained.

### A. Featurization ergonomics — kill the glue
- [ ] **1. (#306) `feature_matrix` straight from `df_seq`.** Let
  `SequenceFeature.feature_matrix` accept `df_seq` + `list_parts` and build
  `df_parts` internally, so the `get_df_parts(...)`-wrapped triple-call
  (notebook cells 12, 14, 18, 24, 30) becomes one call. Keep the existing
  `df_parts=` path working unchanged.
- [ ] **2. (#307) Scale-average baseline featurizer.** Add
[...]

Issue #91, opened 2026-06-06
Model evaluation & comparison: repeated-CV + bootstrap CIs + paired ΔMCC


breimanntools:
## Problem

Going from features to a *rigorously evaluated* model is hand-rolled boilerplate.
Repeated/stratified cross-validation, multi-seed mean±std, bootstrap confidence
intervals, and paired ΔMCC significance were all reimplemented outside AAanalysis
by the antibody-non-specificity project (their `summ()`, `boot_ci()`, 5-seed loop
over seeds 42/17/13/1/22). The package ships only `comp_bootstrap_ci`, and the
existing `.eval` methods are per-class (`CPP.eval`, `TreeModel.eval`) and tied to
feature-set quality — there is no general model-evaluation harness. AAanalysis's
stated stance is "MCC + external hold-out + honest CIs," yet it offers no
turn-key way to produce exactly that, and the rigor matters: in that project,
dPULearn looked like +0.16 MCC on a single hold-out but −0.07 under repeated CV.

## Goal

Provide a first-class evaluation surface for repeated-CV + confidence intervals +
paired model comparison, reusing `comp_bootstrap_ci` and `aaanalysis.metrics`,
honoring the `random_state`/`seed` reproducibility contract, with no new required
dependency (scikit-learn is already core).

## Design options (decide during triage — pick one)

1. **Standalone helpers** — `aa.evaluate(model, X, y, cv=..., metrics=[...], ci="bootstrap")`
   + `aa.compare(model_a, model_b, ...)`. Matches the `comp_*` functional style;
   lowest ceremony; **two new top-level exports (CONFIRM-FIRST)**.
2. **Dedicated class on the `Tool` template (recommended)** — e.g. `ModelEvaluator`
   with `.run`/`.eval` holding per-fold/per-seed results, paired with
   `ModelEvaluatorPlot` (CI bars, ΔMCC) per the 1:1 logic/plot convention. Most
   consistent with the package architecture; stateful and plottable.
3. **Hybrid** — thin `aa.evaluate`/`aa.compare` helpers backed by the class engine
   internally (functional facade + class engine).

## Requirements

- [ ] Repeated / stratified cross-validation over a fitted estimator + `(X, y)`.
- [ ] Multi-seed aggregation (mean±std) honoring the `random_state`/`seed` contract.
- [ ] Bootstrap confidence intervals via `comp_bootstrap_ci`.
- [ ] Paired ΔMCC between two models with a significance test.
[...]

PR #315, opened 2026-06-30
feat: aa.eval_features — model+CV+metric feature-set scorer (prototype #309)

breimanntools:
**Status: SOLID** — works, green local tests, full ripple done. Part of #305 / prototype for #309.

## What to review
- New public scorer `aa.eval_features(X, labels, model=None, cv=None, metric="balanced_accuracy", mask_known_pos=None, random_state=None)` returning a percentage score.
- The **default** (linear SVM + `LeaveOneOut` + balanced accuracy) reproduces the γ-secretase notebook's hand-rolled `balanced_acc` helper exactly on the same feature matrix (golden test asserts `isclose` vs the manual `cross_val_predict(SVC(kernel="linear"), ...)` recipe).
- The PU **`mask_known_pos`** variant (port of `cv_leave_one_out_masked` from the original project's `scripts/ml_benchmarking.py`): masked known positives stay in every training fold but are never scored as test points.

## Design notes / open API questions
1. **Percentage scaling.** Returns the bounded metric × 100 uniformly (matches the notebook's `* 100`). For `matthews_corrcoef` (range [-1, 1]) this yields [-100, 100] — documented, but a maintainer may prefer to restrict the metric registry to [0, 1] metrics, or add a `percentage=True` toggle.
2. **Masked-CV predictions.** The port uses `model.predict` (not the original's `predict_proba > 0.5`) to stay deterministic and avoid requiring `probability=True` on SVC (Platt scaling is stochastic). For a linear SVM these agree on the decision boundary. The default (unmasked) path uses `cross_val_predict` to byte-match the notebook.
3. **Metric registry** is a small fixed dict (`balanced_accuracy`, `accuracy`, `f1`, `precision`, `recall`, `matthews_corrcoef`). Could be widened to all `sklearn.metrics.get_scorer_names()` if desired.
4. **Coordinate with #91 / #310** so the eval/comparison surface is one API (the eval heatmap consumes scores like these).

## Ripple done
- numpydoc docstring with named `Returns` + `.. include:: examples/eval_features.rst`.
- Executed example notebook `examples/metrics/eval_features.ipynb` (default / model+cv+metric swap / PU mask; `display_df` table + embedded outputs).
- 29 unit tests (`tests/unit/metrics_tests/test_eval_features.py`): positive + negative per parameter, reproducibility (same `random_state` → identical score), golden vs manual recipe. All green locally.
- Wired into `__init__`/`__all__` (CONFIRM-FIRST; #309 is on the epic wire-to-public-API list), `metrics/__init__.py`, `docs/source/api.rst` autosummary, release notes (Unreleased · Metrics), cheat sheet.
- Meta-tests green: `test_param_coverage`, `test_utils_barrel`, `test_backend_import_hygiene`, full `api_tests` (175) + `metrics_tests` (112).

## KPIs (#309)
- [x] Default reproduces notebook bACC numbers within tolerance (golden test).
- [x] Same `random_state` → identical score.
- [x] ≥1 unit test per parameter, positive + negative.
- [x] No new dependency (sklearn already a dep).

---

## Critical self-review

Adversarial review pass (not a rubber-stamp). Defects found and fixed:

1. **BUG — `mask_known_pos` + integer `cv` crashed.** The masked path called
   `cv.split(...)` on the raw argument, so `eval_features(X, y, cv=5, mask_known_pos=mask)`
   raised `AttributeError: 'int' object has no attribute 'split'` (reproduced concretely).
   Fixed by normalizing `cv` once with `sklearn.model_selection.check_cv`
   (`None -> LeaveOneOut`, `int -> StratifiedKFold`, splitter passed through) so both the
   default and masked paths get a real splitter. The default/int paths stay byte-identical
   (this is exactly what `cross_val_predict` does internally). Added a `mask + int cv`
   regression test.

2. **Percentage/MCC inconsistency (the flagged design question) — RESOLVED by restricting
   the registry to `[0, 1]` metrics.** `matthews_corrcoef` (`[-1, 1]`) made the `*100`
   scaling land in `[-100, 100]`, breaking the "percentage" contract. I removed it, so every
   registry metric (`balanced_accuracy`, `accuracy`, `f1`, `precision`, `recall`) is bounded
   in `[0, 1]` and the returned score is a genuine percentage in `[0, 100]`. This is the
   smallest, most honest fix and keeps a single return semantics; a signed metric can be
   re-added later behind a scaling-aware design if a use case appears. Docstring, `Returns`,
   and the supported-metric parametrization updated accordingly.

3. **Golden test was near-tautological.** It re-ran `balanced_accuracy_score` against itself.
   Added an independent hand-rolled balanced accuracy (mean per-class recall) as a second
   cross-check, so the assertion actually pins the metric semantics, not just the sklearn call.

4. **PU-mask invariant was only checked indirectly.** Added a rigorous test with a recording
   estimator (class-level storage survives `sklearn.clone`) that asserts the masked known
   positives train **every** scored fold and are **never** handed to `predict` (scored).

5. **Convention/hygiene.** Moved the mid-file `sklearn` imports in `_utils/metrics.py` to the
   module top. Moved the all-samples-masked rejection into the frontend Validate block
   (backend now trusts the frontend), keeping a defensive backend guard — with a clearer
   message — only for exotic custom splitters that never cover an unmasked sample.

Verified: `random_state` is forwarded only to estimators exposing it (guarded by
`get_params()`), so the default SVM and param-less estimators never crash; same
`random_state` -> identical score. No new dependency, no `print`, bare `ValueError` only.
Tests: `metrics_tests` + `api_tests` green (288), example notebook re-executes under `nbmake`,
docstring checker clean for `eval_features` (only a pre-existing `comp_kld` advisory remains).

**Residual note:** the golden test runs on a synthetic `make_classification` matrix, not the
actual γ-secretase feature matrix (that data isn't a lightweight unit-test fixture); it proves
the default reproduces the notebook *recipe* exactly, which is the testable contract.

## Iterative review log
- Round 2 (correctness & edge cases): Found + fixed a real PU-mask bug — with a non-LOO splitter (e.g. `StratifiedKFold`), a masked known positive that landed in a test partition was dropped from **both** scoring and that fold's training, violating the documented "known positives still inform every training fold" contract (concretely reproduced: 3 of 4 folds were missing masked positives from training). Fix folds the masked test indices back into `train_full` per fold; LOO stays byte-identical (single-sample masked test folds are skipped outright, so the default path and golden/notebook reproduction are unchanged). Added a k-fold masked-invariant regression test (31 tests). Also verified single-class labels raise a friendly `check_labels` ValueError, not a cryptic sklearn error.
- Round 3 (numerical robustness & API consistency): Verified adversarially that all five registry metrics return a genuine percentage in [0, 100] even on degenerate/near-random fixtures that trigger zero-division folds, and that the default + RandomForest paths are deterministic under `random_state`. One API-consistency wart: `f1`/`precision`/`recall` (binary-averaged) raise a cryptic sklearn `ValueError` on multi-class labels while `balanced_accuracy`/`accuracy` accept them. It already fails loud (a `ValueError`), so the slim honest fix is documentation, not cross-layer validation plumbing: clarified in the `metric` docstring which metrics are multi-class vs binary, and added a positive multi-class test for the supported metrics (33 tests). No behavior change.
- Round 4 (efficiency & simplicity): Dropped two needless per-fold re-wraps in the masked-CV loop (`np.asarray(y)[keep]` -> `y[keep]`, `np.asarray(preds)` -> `preds`; both are already ndarrays), a recompute-inside-loop the round-4 lens targets — output byte-identical (33 tests green). Audited the rest (frontend Validate block, `eval_features_`, the `.astype(int)` safety cast, default `cross_val_predict` path): no dead code, no redundant validation, no other in-loop recompute, and the per-fold refit loop is inherently non-vectorizable. Otherwise converged.
- Round 5 (guides, docs & tests completeness): Fixed one convention item — `metric` validation now reads the registry via `ut.LIST_METRICS_EVAL` (barrel access) instead of a bare top-level import, matching the house `ut.X` constant rule. Verified: numpydoc named `Returns` + one-line summary; `.. include:: examples/eval_features.rst` backed by a present, nbmake-passing `examples/metrics/eval_features.ipynb` (same generation pattern as the sibling v1.1 metrics); no citations to resolve; no ADR refs introduced by this PR; frontend Validate covers every public param and the backend trusts it; bare `ValueError` only, no `print`. Every param has a positive + negative test. Docstring checker + doc/signature-drift clean for `eval_features` (only the pre-existing `comp_kld` RAISES advisory remains). Full local gate green: metrics_tests + api_tests (param_coverage, backend_import_hygiene, utils_barrel, return-contract) = 301 passed.


breimanntools:
Closing as part of the prediction-layer consolidation. `aa.eval_features` is redundant: `CPP.eval` already scores feature sets by intrinsic quality (discriminative power + redundancy), and the model+cross-validation scoring for grid search already lives inside `find_features`/`CPPGrid` and `AAPred.eval`. No new standalone scorer is needed. Issue #309 stays open until the consolidated capability lands; the branch is retained (not deleted) so this can be revived if needed.

Issue #307, opened 2026-06-30
feat: SequenceFeature.scale_mean — scale-average baseline featurizer


breimanntools:
Part of #305.

## Problem

The standard "scale-based vs CPP" comparison baseline — averaging each scale
over a whole sequence (no positional split) — has no library API, so the
γ-secretase notebook (cell 27) hand-rolls it with a raw residue comprehension:

```python
def scale_X(df):
    seqs = (df["jmd_n"] + df["tmd"] + df["jmd_c"]).to_list()
    return np.array([df_scales_red.loc[[a for a in s if a in df_scales_red.index]].mean(axis=0).values
                     for s in seqs])
```

This is a legitimate, commonly-needed baseline featurization, but every user
reinvents it (and gets the missing-residue filtering subtly wrong).

## Goal

Add a first-class scale-average featurizer that turns sequences + scales into a
`(n_seq, n_scales)` matrix, with the name matching its output noun.

## Requirements
- [ ] Add `SequenceFeature.scale_mean(df_seq, df_scales, list_parts=None)`
      (or `NumericalFeature` if a better fit) returning the per-sequence scale
      average matrix. `list_parts=None` → whole sequence.
- [ ] Handle non-canonical / missing-in-scale residues consistently with the
      rest of the package; document the rule.
- [ ] Optional `return_df=True` for a labeled DataFrame (matches the package's
      `return_df` convention).
- [ ] numpydoc + per-method `Examples` include.

## KPIs / Acceptance criteria
- [ ] Output equals the notebook's manual comprehension on the canonical fixture
      (within float tolerance).
- [ ] ≥1 unit test per parameter (positive + negative); covers an empty/all-
      non-canonical sequence edge case.

## Scope / non-goals
- A no-positional-split mean only; positional splits remain CPP's job.

## Dependencies
- Part of epic #305 · relates #87 (CPP strategy preset).

## Standards checklist
- [ ] frontend/backend · validation block · numpydoc · tests · no-print

breimanntools:
Shipped on `master` as `SequenceFeature.scale_composition(...)` (`aaanalysis/feature_engineering/_sequence_feature.py`), the renamed form of the proposed `scale_mean` (name matches output; PR #319). The scale-average baseline featurizer is available and used by the γ-secretase Use Case. Closing as completed per the epic #305 reconciliation audit.

PR #319, opened 2026-06-30
feat(sf): SequenceFeature.scale_composition — scale-based baseline featurizer (#307)

breimanntools:
Adds **`SequenceFeature.scale_composition`** — a no-positional-split baseline featurizer. For each sequence it averages every scale over the residues of a span, giving the `(n_seq, n_scales)` matrix `X`: the sequence's mean profile in scale-space (the scale-based analogue of amino-acid composition), with no positional information.

**Application.** Build a baseline feature set for a prediction model and compare the same classifier on this `X` versus a `CPP` `feature_matrix`, to show how much the positional Part-Split-Scale features add over a plain scale average (the "scale baseline vs CPP" comparison). It is *not* a positional feature set.

Notes:
- Named `scale_composition` (not `scale_mean`) to fit the compositional-descriptor family — see the follow-up issue #335 (AAC, ACC, and baseline comparison folded into `AAPred`) — and to avoid reading like `df_scales` (the letters × scales matrix).
- Fully vectorized backend (byte→scale lookup + one `bincount` + a single BLAS matmul; no per-sequence Python loop). Correct missing/non-canonical-residue dropping; all-NaN rows for empty spans with a verbose warning. `return_df=True` for a labeled frame.
- `list_parts=None` → whole `jmd_n` + `tmd` + `jmd_c` span.

Tests + api-meta green (195); example notebook re-executed with outputs.


breimanntools (commented):
Why do we need this? This is alreday done in CPP.run(split_n_max=1). However, this is indeed a simplificaiton of the process without filtering redudant features. So, it is worht integrateing. Okay from my side

breimanntools (commented):
This is more a baseline prediction method, or? Should we simply make a baselin predicto and integrate this there? Then we can also add AAC, DPC, CKSAAP and other approaches to get simple baseline predictors? Perhaps in AAPred we have a baseline method. This can load the repreaentations from iFeature +

Issue #308, opened 2026-06-30
feat: dPULearn pos/unlabelled mined-mask convenience + get_labels + named sample colors


breimanntools:
Part of #305.

## Problem

Using `dPULearn` to mine reliable negatives currently forces the user to pack a
combined matrix, build a `1`/`2` label vector, fit, then slice the mined rows
back out by index — repeated in notebook cells 18 and 24:

```python
X = np.vstack([X_cpp_pos, X_cpp_oth])
y = np.array([1]*len(X_cpp_pos) + [2]*len(X_cpp_oth))
dpul = aa.dPULearn(...).fit(X, labels=y)
mined = np.asarray(dpul.labels_)[len(X_cpp_pos):] == 0
```

The same notebook also writes `(df["label"] == x).astype(int).to_numpy()` in
**4+ places** and pulls canonical sample colors out of a string-keyed dict
(`aa.plot_get_cdict(...)["SAMPLES_POS"]`) in cells 6 and 22.

## Goal

Make the positive/unlabelled → mined-negatives flow a single call, and remove
the recurring label-vector and color-lookup plumbing — additively, leaving the
current `dPULearn.fit(X, labels=...)` API unchanged.

## Requirements
- [ ] A `dPULearn` convenience accepting `X_pos` + `X_unlabelled` separately and
      returning the mined reliable-negative **boolean mask** (and/or subframe),
      so users don't `vstack` + label `1`/`2` + slice by hand.
- [ ] `aa.get_labels(df, positive_label=1)` returning the binary `int` label
      vector for the `(df["label"]==x).astype(int).to_numpy()` pattern.
- [ ] Expose the canonical sample colors as named constants (e.g.
      `aa.COLOR_SAMPLES_POS/NEG/UNL/REL_NEG`) or `plot_get_clist(name="samples")`,
      so users stop indexing `plot_get_cdict(...)` by string key.

## KPIs / Acceptance criteria
- [ ] The mined mask equals the current `labels_[len(X_pos):]==0` result exactly
      on the canonical fixture (regression).
- [ ] `get_labels` matches the manual expression on ≥2 label encodings.
- [ ] Named color constants equal today's `plot_get_cdict` values (golden test).

## Scope / non-goals
- No change to the dPULearn algorithm or its existing `fit` signature.

## Dependencies
- Part of epic #305.

## Standards checklist
- [ ] frontend/backend · validation block · CONFIRM-FIRST (new public symbols →
[...]

PR #321, opened 2026-06-30
feat(pu): dPULearn.fit split input (X_pos/X_unlabeled + mask_neg_) + aa.get_labels + sample colors

breimanntools:
Part of #308. Three additive pieces for the positive/unlabeled workflow.

## 1. `dPULearn.fit` — positives/unlabeled split input + `mask_neg_`
For the common positives-vs-unlabeled setup, `fit` now accepts **`X_pos` + `X_unlabeled`** as an alternative to `X` + `labels`: the two matrices are stacked and marked (`1`/`2`) internally, so you don't hand-build the label vector. After fitting, the new **`dPULearn.mask_neg_`** attribute is the boolean mask of reliable negatives — over the rows of `X_unlabeled` in the split mode, over `X` otherwise (equals the manual `labels_[len(X_pos):] == 0` exactly).

`fit` still **returns `self`** (scikit-learn contract preserved — no output-flag anti-pattern), and the existing `fit(X, labels=...)` path is byte-identical.

```python
dpul.fit(X_pos=X_pos, X_unlabeled=X_pool, n_neg=49)
X_neg = X_pool[dpul.mask_neg_]
```

> Supersedes the earlier standalone `dPULearn.mine_negatives` method (removed): the convenience now lives on `fit` itself rather than adding a second entry point.

## 2. `aa.get_labels(df, positive_label=1, col_label="label")`
One-call binary label vector for the recurring `(df[col]==x).astype(int)` pattern.

## 3. `COLOR_SAMPLES_*` constants
Canonical `aa.COLOR_SAMPLES_POS/NEG/UNL/REL_NEG` colors so notebooks reference a named constant.

Tests cover the split-mode ↔ manual equivalence, `mask_neg_` semantics in both modes, the returns-self + both-modes-rejected guards, and `get_labels` edge cases. 271 dpulearn + api-meta tests green.


breimanntools (commented):
Looks good. Please make sure that the alternative is cleraly described in the example notebook and both options are clearly introduced as two options (perhaps bullet points in the beginning of the example notebook before each paratmer is introduced iteslef!

Issue #313, opened 2026-06-30
feat: ShapModelPlot.clustermap (explanation-similarity) + shap_to_feat_imp + sample= SHAP plots


breimanntools:
Part of #305.

## Problem

The richest explanation figure in the original project — clustering samples by
**explanation similarity** (correlation of their per-sample SHAP-value vectors,
so proteins group by *why* the model scores them, not just by features) — is
~130 lines of `sns.clustermap` + `scipy.cluster.hierarchy` surgery
(`plot_shap_cluster_map` / `plot_shap_dendrogram`, `scripts/plot_cpp_shap.py`).
It has no library API. Separately, per-sample SHAP plots in the γ-secretase
notebook (cells 30/32) couple the plot call to `add_feat_impact`'s naming by
hand:

```python
seq_kws = sf.get_seq_kws(df_seq=_df_gsec, df_parts=df_parts_dom, sample=acc)
fig, ax = aa.CPPPlot().feature_map(df_feat=df_feat, shap_plot=True,
                                   col_imp=f"feat_impact_{name}", **seq_kws)
```

The `col_imp=f"feat_impact_{name}"` string-templating + `get_seq_kws` plumbing is
repeated per sample.

## Goal

Ship a SHAP-correlation clustermap and let the existing SHAP plot paths resolve a
sample by name, removing both the 130-line clustermap boilerplate and the
per-sample string plumbing. All SHAP pieces stay in the **`pro`** extra.

## Requirements
- [ ] `ShapModelPlot.clustermap(...)` — correlation clustermap of per-sample SHAP
      vectors with row/col class-color sidebars (port `plot_shap_cluster_map`).
- [ ] `shap_to_feat_imp` helper: SHAP vector → normalized signed feature impact
      (`shap/Σ|shap|*100`) + absolute importance (port `ml_shap.py`); reuse any
      existing `ShapModel` logic to avoid divergence.
- [ ] Let `feature_map` / `ranking` / `profile` accept `sample="APP"` to resolve
      the `feat_impact_*` column + `seq_kws` internally (removes the
      `get_seq_kws` + `col_imp` plumbing from cells 30/32).
- [ ] numpydoc + `Examples` include; `pro`-gated import.

## KPIs / Acceptance criteria
- [ ] `clustermap` reproduces the project figure's clustering on the canonical
      SHAP matrix (linkage + cluster assignment match).
- [ ] `feature_map(..., sample="APP")` equals today's explicit
[...]

PR #323, opened 2026-07-01
feat(pro): shap_to_feat_imp + CPPPlot sample= (SHAP per-sample plumbing)

breimanntools:
Part of #305 / #313, refreshed onto current master.

## What this adds
- **`shap_to_feat_imp`** (pro) — convert a per-sample SHAP vector into normalized signed feature impact / absolute importance, reusing `ShapModel.add_feat_impact`'s backend so the two never diverge.
- **`CPPPlot.ranking` / `profile` / `feature_map` `sample=`** — resolve a sample by name (feature-impact column + TMD-JMD parts) so per-sample SHAP plots need no manual `col_imp=f"feat_impact_{name}"` string plumbing.

## Consolidation note
The explanation-similarity **clustermap moved to the core `AAPredPlot.clustermap`** (drawing from provided importance vectors), so `ShapModelPlot` (which only held the clustermap + its dendrogram cut) was removed here along with its backend and tests. This PR now carries only the SHAP per-sample helpers above.

Part of #313 (no closing keyword).


breimanntools (commented):
The clustermap is not only for shap values but also for featuers or other numerical represntations. Perhaps we need a plot_clustmarp utils insterad of asigning it to SHapModel. Should we make a general plotting class called AAPlot (aap) and the predction plots can be asigned to this one as well. Or AAPredPlot) I do not know right now

breimanntools (commented):
I want to check this in the example notebook! Make sure that this is capability and shortcut is shown there clarely (like first the way we do it so far, then intro that a shortcut for this exist as follows....

Issue #219, opened 2026-06-13
Dense-plot label overlap gate + global `auto_font` option: auto-optimize plot fonts package-wide (+ auto-size feature_map)


breimanntools:
## Problem
Figures with many rows render **overlapping text**. The clearest offender is
`CPPPlot.feature_map`: its left-hand scale **subcategory row labels** collide as
soon as the figure has more than a handful of rows. At the full AAontology
breadth (74 subcategories) the label column becomes an unreadable black blur;
the overlap is already measurable at ~20 rows. There is **no test** that catches
this, so regressions ship silently and existing committed figures likely already
carry mild overlap.

Investigation notes (verified by rendering real figures, 2026-06-14):
- The visible row labels are **hand-placed `ax.text` artists**, not real y-tick
  labels. Path: `feature_map → plot_feature_map → PlotElements.add_subcat_bars
  → ut.plot_add_bars → _add_text_labels` (`aaanalysis/_utils/utils_plot_elements.py`).
  They render at the current rcParams font size (~10pt); a vestigial 1pt
  `set_yticklabels` copy also exists (ignore it).
- `feature_map` is a **5-axis composite** (heatmap + top importance bars +
  colorbar + cumulative-importance + legend). A naive "any text overlaps any
  text" detector is **unusable** — it flags legitimate layered text. The test
  must **scope to the row-label artists**.
- Reliable bbox measurement requires a **forced render** first
  (`fig.savefig(BytesIO())` or `fig.draw_without_rendering()`); a bare
  `get_window_extent` returns degenerate ~1px heights and reports false "0 overlaps".
- **Separate bug found:** the top importance-bar numeric ticks (`'40'` over `'0'`)
  overlap independently of row count — fix in the same PR or split out.

## Goal
1. A **test gate** that renders dense figures and fails on overlapping
   *row labels* (scoped, low-false-positive).
2. `feature_map` (and the other dense plots — `heatmap`, `ranking`, `profile`)
   **self-adjust** so labels never overlap and never drop below a legibility
   floor: **shrink the label font to a floor (~5–6pt), then grow the figure
   height** when even the floor would collide. Per maintainer steer, route the
   font adjustment through **`plot_settings`** (it scales fonts properly and has
   other benefits) rather than a bespoke per-call knob.

## Requirements
- [ ] Add a reusable test helper (e.g. `tests/unit/plotting_tests/_text_overlap.py`)
      that, given a `fig` and a set of target label strings, force-renders and
[...]

PR #326, opened 2026-07-01
feat(plotting): dense-plot overlap gate + auto_font option + feature_map auto-size (#219)

breimanntools:
Makes the dense CPP composite plots stable across data sizes (issue #219), reworked to hold grid **cell size** constant.

## Mechanism
`auto_font=True` (now the **default**) holds each grid cell at a constant physical size and grows the *figure* with the data, so subcategory labels, position ticks and residue letters stay at a constant, legible font regardless of grid size — no manual `plot_settings(font_scale=...)`. It only ever GROWS the figure (floored at the default), so sparse grids never collapse. `auto_font=False` reproduces the previous fixed-size output (byte-identical). Covers `feature_map`, `heatmap`, `profile`, and `ranking`.

## Testing
- Invariant tests (constant cell size on dense grids, fonts neither shrunk nor enlarged, no overlap, small-grid floor, `figsize=None` in both states, `seq_char_fill` follows auto_font). Full plotting + cpp_plot suites green (666).
- auto_font gallery added to the options example notebook; CPP plot notebooks re-executed under the new default.

Closes nothing (issue #219 stays open for the broader font work); part of #219.


breimanntools (commented):
The plots do not look better. This imrpovement did not work!! Do it again and make a before after comparison to demonstrate me that it works. I want robust outscale methods to show for sequences with flexible length optimal plots!!!

breimanntools (commented):
I am not satisfied with the results. We must take a good care of this PR in the end

Issue #336, opened 2026-07-04
epic: usability gaps surfaced by a real epitope-prediction project


breimanntools:
## Context

Usability gaps surfaced while building a real linear + conformational B-cell **epitope-prediction**
pipeline on **aaanalysis master** (`__version__` still reads `1.0.3`) — using CPP, `aap.find_features`,
`EmbeddingPreprocessor`, `StructurePreprocessor`, `ShapModel`, `CPPPlot`, `dPULearn`. Grouping the
findings here; the sub-issues below are the concrete, independently-shippable subtasks.

### Already tracked (this project confirms the need in practice)
- **#335** compositional baseline featurizers (AAC/DPC/…) — *the biggest source of glue code*; had to
  hand-build a 20×20 one-hot **identity** scale set (`Segment(1,1)` = AAC) and a 400-token **dipeptide**
  alphabet for DPC. Not on master yet.
- **#309** `aa.eval_features` — re-implemented a `cross_val_predict`+metric wrapper in ~every script.
- **#306** (closed) `feature_matrix(df_seq=…)` — confirmed present on master, resolved. 🎉
- **#86 / #87** compositional-vs-positional guide — this project is a clean worked example (composition
  wins linear Z1; positional CPP + 3D structure wins conformational Z2).
- **#119 / #280 / #284 / #288** structure + feature-map linked explorer — exactly the app I hand-built.

### New subtasks (not yet tracked)
- [ ] #337 — B1: `NumericalFeature.feature_matrix` for numeric CPP (`run_num`) outputs
- [ ] #338 — B2: `find_features` / Pattern splits fail on free peptides (no flanks)
- [ ] #339 — B3: `CPP.run(n_jobs>1)` `mp.Manager()` `EOFError` outside notebooks
- [ ] #340 — B4: `StructurePreprocessor` per-feature failure isolation + `msms`/backend routing
- [ ] #341 — B5: CPP filter blind to distributed (jointly-strong) signals

### Minor
- `__version__` stays `1.0.3` on master (unbumped) → master install is indistinguishable from the PyPI
  release; a `.devN`/git-hash suffix would disambiguate.
- Discoverability: re-implemented things that already exist (`SequenceFeature.get_labels_quantile/
  get_labels_tiered`, `prune_by_correlation/variance`, `get_df_parts_from_windows`) — a task→method
  cookbook would help (docs epic #106).

breimanntools:
[...]

PR #347, opened 2026-07-04
feat(num_feat): add NumericalFeature.feature_matrix (#337)

### LINKED ISSUE #337: feat: NumericalFeature.feature_matrix for numeric CPP (run_num) outputs
Part of #336 (usability epic).

## Problem
Numeric CPP on embeddings/structure works for **discovery** but not for **use**:
`EmbeddingPreprocessor`/`StructurePreprocessor` + `NumericalFeature.get_parts` + `CPP.run_num` return a
`df_feat`, but there is **no `NumericalFeature.feature_matrix`** to turn selected features into a model
matrix. Worse, `run_num`'s `positions` column uses a JMD-offset numbering (e.g. `21,22,…,30` for a
`jmd=0` TMD) that doesn't map to the `(L, D)` array indices, so I couldn't reconstruct feature values
myself. I abandoned `run_num` and fell back to `build_scales → SequenceFeature.feature_matrix`
(per-AA-averaged), which discards the per-residue context that motivates using embeddings/structure.

## Suggestion
- Add `NumericalFeature.feature_matrix(features, dict_num_parts, df_scales=…)` mirroring
  `SequenceFeature.feature_matrix`, **or** let `CPP.run_num(..., return_X=True)` return the matrix.
- Document how `df_feat.positions` from `run_num` maps back to `dict_num_parts` indices.

## Why it matters
Without this, the embedding/structure branches of CPP can *find* interpretable features but can't feed
them to a downstream classifier without bespoke glue.

breimanntools:
Closes #337.

## Summary

Adds `NumericalFeature.feature_matrix(features, dict_num_parts, df_parts, df_scales=..., n_jobs=1)`,
the numerical analog of `SequenceFeature.feature_matrix`: it turns `CPP.run_num`-selected features back
into a model matrix `X`, preserving the per-residue context that per-AA-averaged sequence features
discard. `run_num` returns `df_feat` (selected feature ids + stats), never `X`; this is the missing
materializer that turns those ids into the model matrix — the numerical-mode counterpart of the
sequence-mode `SequenceFeature.feature_matrix` step used throughout the `aapred_*` / `seqopt_*` examples.

## Details

- Values are reconstructed exactly the way `CPP.run_num` does — the SPLIT in each feature id is
  re-applied to the part's 0-based residue axis (`arange(L_part)`), the SCALE selects the column, and
  the selected residues are nanmean-averaged (round 5).
- **Per-part length `L_part` comes from `df_parts`** via the *same* helper `run_num` uses internally
  (`_derive_dict_part_lens`, non-gap character count), not inferred from the tensor's NaN padding. So
  `X` is byte-identical to `run_num`'s value reconstruction in **every** case — including when a
  genuine residue is all-NaN across `D` (an unresolved structure position or masked embedding), which
  a NaN-based length would have mis-counted as padding and shifted the split boundaries. `_cpp.py` is
  untouched, so `run_num` itself is unaffected.
- The `df_feat` `positions` column is a JMD-offset display numbering, not a tensor index, so it is
  deliberately not used for value lookup (documented in the method Notes).
- The frontend validates `df_parts` (row count, part-column coverage, real length ≤ padded tensor
  length) before dispatch.
- Heavy lifting in `_backend/num_feat/feature_matrix.py`. `@staticmethod`, no `__init__.py` change.

## Verification

Byte-identical to `run_num`'s engine (`recompute_feature_matrix`) for uniform, ragged/variable-length,
different-`D`, and the previously-divergent all-NaN-real-residue inputs.

## Ripple

- numpydoc docstring (named Returns / Raises / Examples include)
- executed example notebook `examples/nf_feature_matrix.ipynb` (every public parameter, `display_df` tables)
- unit tests (per-parameter positive+negative including the new `df_parts` arg, golden hand-computed
  means, `run_num` consistency incl. all-NaN-real-residue parity, ragged parts)
- release-notes Unreleased entry

Part of epic #336.


breimanntools (commented):
Is this really necesary? dict_num_parts goes into cpp_run__num and we then get X and not dict_num_parst to X. Why did you add this here? Was there any use caes that required this

Issue #339, opened 2026-07-04
CPP.run(n_jobs>1) crashes with multiprocessing.Manager() EOFError outside notebooks


breimanntools:
Part of #336 (usability epic).

## Problem
`CPP.run(..., n_jobs>1)` crashes in non-notebook contexts (`python -c`, heredoc, some subprocess
shells). The progress-sharing path `.../cpp/_filters/_progress.py:_get_mp_shared()` calls
`multiprocessing.Manager()`, which raises:

```
EOFError
  File ".../multiprocessing/managers.py", start -> reader.recv() -> _recv_bytes -> raise EOFError
```

Workaround: `n_jobs=1`, or always run from a real `.py` module file. This makes multi-core CPP fragile
in CI/scripts/pipelines.

## Suggestion
Wrap the `mp.Manager()` spawn in try/except and **degrade gracefully to single-process progress**
(or a no-op progress bar) instead of aborting the run.

PR #349, opened 2026-07-04
fix(cpp): make CPP splits usable on free peptides with no flanks (#338)

### LINKED ISSUE #338: find_features / Pattern splits fail on free peptides (no flanks)
Part of #336 (usability epic).

## Problem
`find_features` (and CPP `Pattern`/`PeriodicPattern` splits) are unusable on **free peptides with no
flanking context** — the entire linear-epitope use case. `Pattern`/`PeriodicPattern` carry a hardcoded
`len_max=15`; the `find_features` search stage applies them regardless of `kws["n_split_max"]`, and it
carves a JMD out of the sequence, so a short free peptide (target <15 aa after JMD) raises:

```
ValueError: For split_kws (n_max=15): '{'Segment': ..., 'Pattern': {... 'len_max': 15}, ...}',
  following 'tmd' part contains too short sequences (e.g., 'PQFTIFGT', n=8).
```

The message doesn't explain the real cause or the fix, and it recurs for any peptide dataset shorter
than ~15 aa in the target region.

## Suggestion
- Auto-drop `Pattern`/`PeriodicPattern` when a part is shorter than their `len_max` (or when `n_jmd=0`),
  and honor `kws` `len_max`.
- Raise a clear, actionable error: *"target region too short for Pattern splits — use Segment-only or
  add flanks."*
- Consider a documented "free peptide / no-flank" recipe for `find_features`.

breimanntools:
Closes #338.

## Summary

`Pattern`/`PeriodicPattern` splits and `aap.find_features` were unusable on free peptides with no
flanking context (the linear-epitope case): the default split config requires each part to be >= 15
residues, so any target region shorter than ~15 aa raised an opaque "too short" `ValueError`.

## Details

- **Actionable error**: `check_match_df_parts_split_kws` now names the binding split length (which
  split type / parameter drives `n_max`) and the concrete fix (Segment-only splits, lower
  `len_max`/`n_split_max`, or add `jmd_n`/`jmd_c` context).
- **Honor kws**: `find_features` `kws` now accepts `len_max` and actually threads
  `n_split_max`/`len_max` into the split config for the fast path (previously ignored) and the
  `CPPGrid` search stages.
- **Auto-fit (fast path)**: the split config auto-fits to the shortest part, dropping
  `Pattern`/`PeriodicPattern` and clamping `n_split_max` with a single `UserWarning`, so free peptides
  run out of the box. Byte-identical when parts are long enough.
- **Search Stage 3**: the simplify CPP now uses the winner's `split_kws` instead of the default,
  which previously hard-errored on short parts even though the grid had gracefully soft-dropped the
  non-fitting configs.

Results for flanked inputs are unchanged (byte-identical), proven by a fast-path A/B hash comparison
master-vs-branch.

## Ripple

- unit tests: actionable message, Segment-only / reduced-`len_max` short-part paths, the auto-fit
  helper (drop/clamp + warning), and free-peptide fast/balanced runs.
- release-notes Unreleased Fixed entry, placed in the Version 1.1 section (consolidates with #339's
  Fixed entry on merge).

Part of epic #336.


breimanntools (commented):
Free peptides work if we asssign the complete peptide to TMD. What was not working there? JMD len=0 is then used!

PR #359, opened 2026-07-06
feat: SequencePreprocessor.pad_parts for uniform gap-padding of df_parts

breimanntools:
## Summary
Adds `SequencePreprocessor.pad_parts(df_parts, length=None, gap="-", pad_at="C"|"N"|"both", list_parts=None)` — pads the selected part column(s) of a `df_parts` DataFrame to a uniform length with a gap symbol.

- `length=None` → each selected column is padded to **its own max** string length (per-part); an explicit `length` pads to that length (raises if a value is longer).
- `list_parts` selects which column(s) to pad (default = all); the others are returned unchanged.
- `pad_at`: `C` (end), `N` (start), or `both` (symmetric — `floor(k/2)` at N, remainder at C).
- Returns a **copy**; the input `df_parts` is never mutated; index and column order preserved.

## Why
Enables the recipe **`pad_parts` → `CPP(accept_gaps=True)` → uniform `n_split_max`**: a padded part has a longer string length, so CPP's split validation permits a finer, uniform `n_split_max` than the shortest *real* part would allow (useful for free peptides / linear epitopes / variable-length parts). The CPP too-short auto-cap warning (from #338) now points users at `pad_parts`. Caveat, documented: padded positions carry the gap-scale value and so feed into features — pad only as far as needed.

## No CPP behavior change
The only `_cpp.py` edit is the auto-cap **warning text** + an `accept_gaps` **docstring** pointer — no logic. Verified: the exact-value CPP **regression anchor passes**, and `encode_one_hot`/`encode_integer` are **byte-identical** (the backend `pad_sequences` gained a `length` param that defaults to the previous pad-to-max behavior).

## Tests & docs
- `pad_parts` — **33 tests** across 9 classes (input / length / gap / pad_at / list_parts / invariants / both / golden references + an `encode_*` byte-identity guard) incl. 6 hypothesis property tests.
- Narrated example notebook `examples/data_handling/sp_pad_parts.ipynb` (a section per parameter + a recipe section), executed with embedded outputs.
- numpydoc docstring (Application / Parameters / Returns / Raises / See Also); release-note entry.

Part of epic #336; follow-up to #338.


aaanalysis/data_handling/_seq_preproc.py
```diff
@@ -239,6 +241,99 @@ def encode_integer(list_seq: Union[List[str], str],
...
+    # Sequence padding
+    @staticmethod
+    def pad_parts(df_parts: pd.DataFrame,
+                  length: Optional[int] = None,
+                  gap: str = "-",
+                  pad_at: Literal["C", "N", "both"] = "C",
+                  list_parts: Optional[Union[List[str], str]] = None,
+                  ) -> pd.DataFrame:
+        """
+        Pad sequence-part columns of a ``df_parts`` DataFrame to a uniform length with a gap symbol.
```
breimanntools, line 253: Do not use the word (pad) to explain it. FInd another expalatin for it (please keep this as well as a rule or epxlain the word in teh docstring at least

breimanntools (commented):
Please improe docstring and push to maser

Issue #335, opened 2026-07-03
Compositional baseline featurizers (scale, AAC, ACC, …) + baseline comparison in AAPred


breimanntools:
## Problem

CPP produces positional Part–Split–Scale features. To show that CPP's positional
features actually beat simpler encodings, users need **baseline featurizers** — fixed-length,
non-positional sequence descriptors — to compare against (the use case's
"scale 61/60/75 vs CPP 71/74/94" barplot). Today:

- Only the scale-average baseline exists (`SequenceFeature.scale_composition`, née
  `scale_mean`). Other standard compositional/statistical descriptors are missing.
- Getting a baseline number is a **manual in-between step**: `df_seq → X_baseline →
  model → score`. The baseline is not a first-class capability of the class that does
  prediction/evaluation (`AAPred`), so every comparison re-implements the loop.

## Goal

1. A small, consistent **family of sequence-level baseline featurizers** on
   `SequenceFeature` — each `df_seq → X` with no CPP positional split — covering the
   common descriptors used as ML baselines for protein sequences.
2. **Baseline comparison built into the prediction class**: `AAPred.eval(...,
   baseline=...)` computes the chosen baseline `X` internally from `df_seq` and reports
   its score as an extra row of `df_eval`, next to the CPP-features score — so the whole
   "CPP vs baseline" comparison (and `aap.plot_eval` barplot) comes from one call, with
   no manual featurization.

## Scope / requirements

**Baseline featurizers** (each: `df_seq` + `list_parts` in, `X` (n_seq, n_dim) out;
`return_df` option; honest signatures; fully vectorized — no per-sequence Python loop;
non-canonical/gap handling documented; ≥30 tests incl. a byte-identical reference-value
test on a fixture):

- **scale-based** — `scale_composition` (mean scale value per scale over the span).
  *Already shipped* (#319); this issue only tracks it as the first family member.
- **amino-acid composition (AAC)** — fraction of each canonical amino acid (20-dim).
- **auto-cross-covariance (ACC)** — lag-based auto/cross-covariance of scale values
  (captures short-range sequence order the pure composition drops); `n_lag` parameter.
- *(candidates, decide during design)* dipeptide composition (DPC), CTD, pseudo-AAC (PAAC).

**Prediction-class integration:**

- `AAPred.eval(df_seq=..., baseline=<kind|True|list>)` builds the baseline `X`
  internally and appends one `df_eval` row per baseline kind, using the same model/CV as
[...]

PR #360, opened 2026-07-06
feat: composition featurizers (kmer_composition) + CPP.run_composit composition mode (#335)

breimanntools:
Advances #335 (does not close it — the ACC/CTD/PAAC descriptors and the AAPred-side comparison remain).

## Summary

Composition featurizers + a composition mode for CPP, so users can quantify how much CPP's positional
Part-Split-Scale features add over plain composition. Two layers, cleanly separated:

**1. `SequenceFeature` — composition featurizers (iFeature-style)**
- `kmer_composition(df_seq, k=1, …)` — one general k-mer-composition featurizer. `k=1` = amino-acid
  composition (AAC, 20-D), `k=2` = dipeptide composition (DPC, 400-D), higher `k` = general k-mer.
  The existing `aa_composition` / `dipeptide_composition` delegate to it (byte-identical); one
  vectorized backend (`get_kmer_composition_`).
- `return_scales=True` → also returns the CPP-ready `(df_scales, df_cat)`: for `k=1` the one-hot
  identity scale set + AA-class `df_cat`; for `k≥2` `df_scales=None` + a k-mer `df_cat`.
- Docstring now cites **iFeature** (Chen et al. 2018) and explains the compositional approaches
  (AAC / DPC / k-mer; CTD / PAAC / DDE noted as parity follow-ups).

**2. `CPP.run_composit(composition="aac"|"dpc"|"kmer", k=…)` → composition `df_feat`** (+ `run_aac`)
- **AAC** *is* positional CPP — a one-hot identity scale set with the whole-part `Segment(1,1)` split
  run through the full pipeline, so it yields real `<PART>-Segment(1,1)-<AA>` features **with
  positions** that `CPPPlot.feature_map` draws.
- **DPC / k-mer** are **non-positional** (a k-mer is an adjacent-tuple property, not a per-residue
  scale): the `20**k` k-mers are scored with CPP's own discriminative statistics (`add_stat_`:
  adjusted AUC, mean-dif, std) and filtered by AUC (top `n_filter`), a min-occurrence guard
  (`min_count`), and optional correlation dedup (`max_cor`). No `positions` (not feature-map-able).

The **CPP-vs-composition comparison** is now simply "produce a composition `df_feat` via `run_composit`
and compare `df_feat`s" — it is *not* baked into `find_features` (that stays pure CPP AutoML).

## Scope note

An earlier iteration wired the comparison into `find_features(baselines=…)`; that coupling was
**removed** in favor of `CPP.run_composit` (the code is preserved in git history for a possible
standalone comparator follow-up). `find_features` is back to its master behavior.

## Ripple

- numpydoc (checker 0-defect, drift 0, param-coverage green); `[Chen18]` added to `references.rst`.
- tests: `test_sf_kmer_composition.py` (incl. `return_scales`) + `test_cpp_run_composit.py` (AAC
  positional / DPC-k-mer non-positional / ranking / min_count / max_cor / validation).
- executed example notebooks: `sf_kmer_composition.ipynb`, `cpp_run_composit.ipynb`.
- release-notes entries.

No `__init__.py` change (methods on the already-exported `SequenceFeature` / `CPP`).


aaanalysis/pipe/_composition_baseline.py
```diff
@@ -0,0 +1,156 @@
+"""
+This is a script for the composition-baseline helpers behind ``find_features(baselines=...)``.
+
```
breimanntools: These composit baseline should also be exposed via the normal API (not only in finder features!

breimanntools (commented):
I want kmers, aac ... also being eposed via the normal API and then shown in the faeture map or an altarnice!

breimanntools (commented):
All good!

PR #364, opened 2026-07-07
docs(shap,cpp): clarify sample-level vs group-level mean_dif for CPP-SHAP maps

breimanntools:
## Summary

Docstring-only clarification of a recurring point of confusion in the CPP-SHAP workflow: a **sample-level** feature map/ranking must be colored by a *sample-specific* difference (one protein minus the reference-group average), **not** the group-level `mean_dif` from `CPP.run` (test group minus reference group, which is identical for every sample).

## Changes (2 files, docstrings only)

- **`ShapModel.add_sample_mean_dif`** — a `.. note::` stating this per-sample column is what a sample-level CPP-SHAP map/ranking should be colored by (pass as `col_val` to `CPPPlot.feature_map`/`ranking`/`profile`), and to choose `label_ref` for the intended contrast.
- **`CPPPlot.feature_map`** — a matching `.. note::` (compute the per-sample column via `ShapModel.add_sample_mean_dif`; reusing the group-level `mean_dif` would show the group signature under every sample), plus a `col_val` parameter clarification.

No code changes; defaults and behavior unchanged. Docstring structural checker passes (0 defects).

### DIFF
--- a/aaanalysis/explainable_ai_pro/_shap_model.py
+++ b/aaanalysis/explainable_ai_pro/_shap_model.py
@@ -814,6 +814,16 @@ class ShapModel(Wrapper):
         The calculated differences are added to ``df_feat`` as new columns named ``mean_dif_'name(s)'``,
         corresponding to each sample or group.
 
+        .. note::
+
+           This per-sample column (sample minus the ``label_ref`` group average) is what a sample-level
+           CPP-SHAP map or ranking should be colored by: pass it as ``col_val`` to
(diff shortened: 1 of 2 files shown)


breimanntools:
Approving — docstring-only clarification, no code/behavior change; docstring structural checker passes. Will merge once the unit matrix + code-quality + integration checks are green (the required A/B benchmark fails on the unrelated 1.0.3-baseline drift).

PR #371, opened 2026-07-08
feat(CPP): bootstrap stability annotation (bootstrap=True) for run/run_num/run_composit; drop redundant run_aac

### LINKED ISSUE #368: feat(CPP): bootstrap / stability feature selection (reference-only resampling, selection-frequency output)
## Summary
Add **bootstrap / stability feature selection** to CPP. Today `CPP.run(...)` does a **single-pass univariate
filter** (`n_filter` on `mean_dif`/`abs_auc`) — there is no resampling-based stability selection. Proposal: resample
the training data B times, re-select each time, rank features by **selection frequency**, and expose that frequency
so users get a **robust, reproducible feature list** (not one fragile to the specific sample). (Note: `comp_bootstrap_ci`
exists only for *score* CIs, not feature selection.)

## Motivation (evidence, Federle 10-benchmark study — `docs/LEAKAGE_STABILITY.md`)
- **CPP's single-pass selection is unstable.** Mean pairwise Jaccard between selections on bootstrap resamples is
  only **0.05–0.22** (worst on low-signal conformational tasks) — a different training sample yields a largely
  different top-K, so any single feature list over-interprets sampling noise (ties to #341).
- **Reference-only resampling is markedly more stable in 10/10 benchmarks** (~1.7–2.8×; mean Jaccard 0.197 vs 0.087).
  Fixing the positives (test group) and resampling **only the reference/negatives** isolates the main instability
  source. ⇒ expose a `resample="both" | "reference"` option.
- **Selecting on the full data leaks** (up to **+0.20 AUC** on DiscoTope). Stability selection must be usable
  **inside a train fold** (leakage-safe), and the docs should say so.

## What bootstrapping does (and does NOT) buy — set expectations
- ✅ **Reproducible, trustworthy feature list + per-feature stability scores** (for interpretation/reporting).
- ❌ **Not higher accuracy**: single top-K ≈ frequency-ranked top-K (±0.002), and OOD transfer is flat across
  2–100 rounds. CPP signal is *distributed* over ~150 weak features, so a broad top-K already captures it; a strict
  "stable-only" threshold (≤5 feats) *hurts*. Document this so users pick top-K-by-frequency, not a hard threshold.
- **Convergence:** ~20–50 rounds is enough (top-150 ranking ~80% converged by 20, ~90% by 50).

## Proposed API
```python
df_feat = aa.CPP(...).run(
    labels=...,
    n_bootstrap=50,              # 0 = current single-pass behaviour (default, back-compat)
    resample="reference",        # "both" | "reference" (positives fixed) | "test"
    bootstrap_frac=0.8,
    n_filter=150,                # rank by selection frequency, keep top n_filter
    random_state=...,
)
# df_feat gains a `selection_frequency` column (0..1); ranking by frequency then mean_dif.
```
- Keep default `n_bootstrap=0` (unchanged behaviour). Reuse the existing `n_jobs` parallelism over rounds.

## Relation to existing issues
- **#341** (CPP filter blind to distributed signals) — complementary: stability selection *reveals* the instability
  that the distributed-signal problem causes, but does not solve distributed detection on its own.
- Pairs naturally with a leakage-safe `eval_features`/CV recipe and #365 (structural CPP) uses the same protocol.

## Acceptance
- `n_bootstrap>0` returns a `df_feat` with a `selection_frequency` column, ranked by frequency; `resample="reference"`
  yields higher round-to-round Jaccard than `"both"` on the benchmark datasets; docs note the stability-vs-accuracy
  trade-off and mandate fold-internal use for leakage-safe CV.

breimanntools:
## Summary

Adds opt-in **bootstrap stability annotation** to CPP (issue #368), and removes the redundant `run_aac`.

`CPP(bootstrap=True)` is a **thin wrapper** over the run methods: it re-runs the ordinary selection on `n_bootstrap` resamples of the data to score how often each feature is selected, then returns the **ordinary full-data run with a `selection_frequency` column** added. The selected features are exactly those of a normal run (`n_filter` stays the selection criterion); bootstrap annotates their stability, it does not change the list.

Closes #368

## API

- Constructor gate + tuned config: `bootstrap=False` (default, byte-identical single pass), `n_bootstrap=20`, `resample="reference"` (or `"both"`/`"test"`), `bootstrap_frac=0.8`.
- Works for **`run`, `run_num`, and `run_composit`** (the wrapper is method-agnostic — per round it constructs a fresh `CPP(bootstrap=False)` on the resample and calls the same public method).
- `df_feat` gains a `selection_frequency` column (0 to 1) when `bootstrap=True`; not combinable with `n_batches`/`n_sample_batches` (raises).
- **`run_aac` removed** — it was a pure alias of `run_composit(composition="aac")`, which fully covers it (unreleased, no deprecation).

## What it is / isn't

- Reports, per selected feature, how reproducible it is under resampling (a trust/interpretability aid).
- Does **not** change which features are selected — the list equals a normal run. (Making the *list* more robust would need candidate-restriction, which was rejected in favour of the simple wrapper — see the ADR.)

## Empirical (DOM_GSEC / gamma-secretase)

- Motivation: single-pass selection is sample-fragile (Jaccard ~0.06 between two data resamples) — so knowing which selected features are reproducible is useful.
- `n_bootstrap` sets the precision of the `selection_frequency` estimate (~20 sweet spot, ~50 converges); `bootstrap_frac=0.8` conventional. A frequency *threshold* over-prunes CPP's distributed signal (0.5 -> 8 features), which is why frequency is annotated, not used to filter (full evidence in the ADR).

## Ripple

Docstrings, `DICT_DF_FEAT` + `_schemas.py` + regenerated `df_schemas.rst`, release notes, `CONTEXT.md`, the re-executed `cpp.ipynb`, the ADR, and the test suite (`test_cpp_bootstrap.py`; `run_aac` test removed from `test_cpp_run_composit.py`).

## Tests / local gates

Bootstrap suite (per-param positive/negative, run/run_num/run_composit, annotate-equals-normal-run, reproducibility, byte-identical-when-off), full CPP suite green, api meta-tests green (param-coverage, notebook param-coverage, df_feat contract, schema doc-sync, class-abbreviation registry, backend-import hygiene), docstring + flake8 clean.


breimanntools (commented):
I thought it would be a simple wrapper around the three CPP run methods and not a method for itself??

PR #369, opened 2026-07-08
feat(prediction): AAPredPlot eval heatmap + confidence-banded predict_group hist

breimanntools:
## What

Two additive `AAPredPlot` plotting modes that collapse recurring glue from the
γ-secretase Use Case notebook (`use_cases/use_case1_gamma_secretase.ipynb`, the
Appendix "prototype convenience API") into one-call APIs. Follows up the (closed)
convenience-methods epic #305 by closing the two remaining **plotting** gaps.

### 1. `AAPredPlot.eval(kind="heatmap")`
Square annotated heatmap of a 2D score grid (`df_eval` rows × columns = the two
sweep axes) with the **optimal cell boxed** and a grid-height colorbar — one call
for the "grid of scores → seaborn heatmap → mark the best configuration" block.
New params: `highlight` (`"max"`/`"min"`/explicit `(row, col)`/`None`), `vmin`,
`vmax`, `cmap`, `cbar_label`.

### 2. `AAPredPlot.predict_group(kind="hist", band=True)`
Color histogram bars by the **confidence band** they fall into (delimited by
`thresholds`) instead of by class `labels` — for scoring **unlabeled** candidates.
Band colors are sampled from `cmap` or taken from an explicit `colors` list.

Both are purely additive: default outputs of every existing `kind` are unchanged.

## Audit (appendix helpers → library)
The Appendix prototypes 5 helpers "that would live in `aaanalysis`". Audited each
against `master`:

| Appendix helper | Status |
|---|---|
| `get_feature_matrix` | ✅ already shipped — `SequenceFeature.feature_matrix(df_seq=, df_parts_kws=)` |
| `PredictionPlot.ranking` | ✅ already shipped — `predict_group(kind="ranking", col_std=, colors=, cutoffs=)` |
| `plot_pca_4groups` | ✅ already shipped — `dPULearn.project` + `dPULearnPlot.pca(df_pu_add=)` |
| `shap_feature_map` | ✅ already shipped — `ShapModel.add_*` + `CPPPlot.feature_map(shap_plot=True)` / `aap.explain_features` |
| **`PredictionPlot.hist`** (confidence-banded) | ❌ **added here** (`band=True`) |
| **`annotated_heatmap`** (box best cell) | ❌ **added here** (`kind="heatmap"`) |

## Also
Fixes a **pre-existing** stale `level="seq"` → `"sequence"` in
`examples/prediction/aapred_plot_group.ipynb` (invalid since the predict-level
[...]

PR #373, opened 2026-07-08
docs(examples): before/after notebook for AAPredPlot convenience plots

breimanntools:
**Stacked on #369** (base = `feat/use-case-convenience-plots`; the notebook calls `AAPredPlot.eval(kind="heatmap")` + `predict_group(band=True)` from that PR). Rebase onto `master` once #369 merges.

## What

A self-contained example notebook (`examples/aapredplot_before_after.ipynb`) that puts the **hand-rolled "before"** code + figure directly above the **one-call "after"** for each prediction-result plot, so the equivalence and the code reduction are visible side by side:

| Figure | Before | After |
|---|---|---|
| Eval heatmap (best cell boxed) | `sns.heatmap` + manual `Rectangle` | `AAPredPlot().eval(kind="heatmap")` |
| Grouped comparison bars | manual `ax.bar` + `ax.text` loop + `axhline` | `AAPredPlot().eval(kind="comparison")` |
| Confidence-banded histogram | `ax.hist` + per-patch recolor | `predict_group(kind="hist", band=True)` |
| Ranked candidates | `barh` + `errorbar` + cutoffs | `predict_group(kind="ranking")` |
| PU-PCA (4 groups) | least-squares affine map + hand scatter | `dPULearn.project(X)` + `dPULearnPlot().pca(df_pu_add=…)` |

Uses small **synthetic** inputs (runs in seconds; the point is the plotting code, not the analysis). A final cell tallies it: **56 → 21 lines of plotting code (~62% fewer)**. Committed with embedded outputs.

## Verification
- Executed end to end (0 errors), every before/after figure renders; visually confirmed the pairs match (heatmap box on the max cell, PCA overlays the projected 4th group, histogram bands blue→red).
- Meta-tests pass with the notebook present (`test_class_abbreviation_registry`, `test_param_coverage`): the only registry-checked assignment is `dpul = aa.dPULearn(...)` (canonical); the `AAPredPlot()`/`dPULearnPlot()` calls are inline.

## Notes
- Placement: `examples/` (it is an example notebook). It is not referenced by a docstring `.. include::`, so Sphinx emits a non-fatal "not in a toctree" warning (`nitpicky=False`, no `-W`) — the RTD build still succeeds. Happy to instead link it from a docstring or move it if you prefer.
- Security: bundled/synthetic data only; no network, file, subprocess, or exec surface.


breimanntools:
### Review + security (done)

**Security:** clean. Inputs are purely **synthetic** (`np.random`), no network / file / subprocess / `pickle` / `exec`; the only `eval(` occurrences are the `AAPredPlot().eval(...)` **method** (not Python's builtin).

**Correctness:** the notebook executes end to end with **0 errors** and every before/after figure renders. Visually confirmed each pair is equivalent (heatmap box on the max cell, PCA overlays the projected 4th group, histogram bands blue→red across the thresholds, ranking bars with error caps). The "before" cells are authentic hand-rolled `matplotlib`/`seaborn` glue; the "after" cells use the real public API.

**Meta-tests:** `test_class_abbreviation_registry` + `test_param_coverage` pass with the notebook present — the only registry-checked instantiation is `dpul = aa.dPULearn(...)` (canonical abbreviation); `AAPredPlot()` / `dPULearnPlot()` are called inline (unchecked).

Left as a **draft** for your review, stacked on #369.

breimanntools (commented):
Only notebooks changed? Show me the final implemetnation!

PR #374, opened 2026-07-08
fix(cpp): stacked cumulative SHAP feature_map bars + heatmap/ranking sizing (#219)

breimanntools:
## Plot-consistency fixes (auto_font review hub, #219)

Three fixes to the CPP composite plots, from the review in #219. Advances #219;
does **not** close it (it stays open as the plot-consistency hub).

### 1. CPP-SHAP `feature_map` bars → cumulative one-direction stacks
The SHAP impact bars had been changed to a **signed-diverging** design (positive
one way, negative the other). Restored to **cumulative stacks in a single
direction**: both the top per-position bars and the right per-subcategory bars now
stack one thin white-edged segment per feature (red = positive, blue = negative),
so no bar extends backwards from the baseline. Test
`test_shap_bars_are_signed_diverging` → `test_shap_bars_are_stacked_one_direction`.

### 2. Heatmap: taller cells (`HEATMAP_CELL_H_IN`)
The standalone `heatmap` grid fills more of its figure than `feature_map`, so at the
shared cell height its rows came out short and the subcategory labels overlapped. A
taller heatmap-specific cell makes the figure a bit higher than wide and gives each
row enough height for its label. `feature_map` cell size is unchanged.

### 3. Ranking: keep the roomy default height
Under `auto_font`, a short ranking was shrinking below the default height. Floor the
auto height at the default (5) so bar spacing stays roomy; taller only when the item
count needs it.

### Validation
- All 352 `tests/unit/cpp_plot_tests` pass.
- `auto_font=False` output is unchanged (all three fixes are inside the
  `auto_font` / `shap_plot` branches).
- **Non-required `visual-regression` will go red** (the heatmap baseline changes by
  design); regenerate baselines via the workflow's `regenerate=true` dispatch after review.

### Notebooks
Example/tutorial notebooks that render these plots are re-executed on this branch so
the RTD build shows the new output. The `use_cases/` γ-secretase notebook is **not**
touched here — it is being edited in a separate session and will be re-run there.


breimanntools (commented):
I need to see it to say if it is good. Push to master and update notebooks

PR #375, opened 2026-07-08
feat(pipe): aap.predict_samples draws the model comparison (plot=True)

breimanntools:
**Stacked on #369** (base = `feat/use-case-convenience-plots`; the plot reuses `AAPredPlot.eval(kind="eval")`). Rebase onto `master` once #369 merges. Part of the **1+2** plan — this is item 2; item 1 (the compact `aap`+`aa` g-sec workflow) builds on this.

## What

`aap.predict_samples` previously returned `(predictors, None, df_eval)` — the middle slot was *always* `None`, explicitly *"reserved for a future comparison plot"*. It's the only pipe function without a figure, while `find_features` / `explain_features` each draw their canonical one. This closes that asymmetry:

- **`plot=True`** (new default) draws the **model comparison** bar plot from the comparison table (hue = model, one bar group per metric, cross-validation `std` error bars) via `AAPredPlot.eval(kind="eval")`, and returns its `Axes` in the middle slot → `(predictors, ax, df_eval)`.
- New style params **`figsize`**, **`dict_color`**, **`baseline`** (chance line).
- Multiple feature sets → bars are labelled `feature_set · model`.
- The **comparison table (`df_eval`) is unchanged** whether or not the plot is drawn — only the previously-`None` slot changes.

Note: `predict_samples` returns a *model comparison* (per-model CV metrics), so its figure is the comparison bar plot — **not** a per-sample score histogram/ranking (those need per-sample scores from `AAPred.predict` + `AAPredPlot.predict_group`, which the g-sec compact workflow uses directly).

## Verification
- New tests: `plot=True` returns `Axes` with the right bar count, `plot=False` returns `None`, table identical with/without plot, `figsize`/`dict_color`/`baseline`, multi-feature-set, invalid `plot`/`baseline`. Existing `test_returns_triple` updated for the new default.
- param-coverage, docstring checker (0 defects), doc-signature drift (0), flake8 `E9,F63,F7,F82` — all green locally.
- Example notebook `examples/pipe/aap_predict_samples.ipynb` re-run (renders the comparison figure); release note added.

No new dependency; no `__init__.py`/`__all__` change (additive params on an existing public function).


breimanntools (commented):
I need to see it. We will go deeper into this step by step later. Now the aap pipelines will be only drafted and improved through more applications

breimanntools:
### Review + security (done)

**Security:** clean. The change is a pure-pandas table reshape (`_comparison_long`) plus a matplotlib draw via `AAPredPlot.eval` — no network / file / subprocess / `pickle` / `exec`; the only `eval(` is the `AAPredPlot().eval(...)` method.

**Adversarial code review** — verified the reshape (metric extraction + order preservation), the `FigAxResult` unpacking (`eval(kind="eval")` always yields a single `Axes`), validation of all four new params, no caller relies on the old `None` slot, and no circular import (`pipe` imports `prediction`, not vice-versa). Findings addressed:
- **[Low]** `dict_color` was silently ineffective in multi-feature-set mode (bars are labelled `"<feature_set> · <model>"`) → documented the composite-label caveat.
- **[Low]** Two tests asserted only `isinstance(ax, Axes)` → now assert the multi-feature-set **bar count (12)** and that **`figsize` took effect**.
- **[Low]** Example notebook didn't exercise `figsize`/`dict_color`/`baseline` → the named-models cell now demonstrates all three.
- **[Low, kept]** `plot=True` draws + returns an open figure by default — this is the **same convention** as `find_features`/`explain_features` (caller owns the figure), so kept for consistency.
- **[Low, negligible]** Label collision only if a feature-set name literally contains `" · "` (U+00B7); feature-set/model keys are dedup-unique and never do — left as-is.

**Gates (local):** `TestPredictSamplesPlot` 7/7, param-coverage, docstring checker (0 defects), doc-signature drift (0), flake8, nbmake — all green. CI here runs RTD only (base is the #369 branch, not `master`); the full matrix runs once retargeted to `master`.

Draft, stacked on #369.

PR #370, opened 2026-07-08
docs(use-case): re-author g-sec notebook onto AAPredPlot convenience methods

breimanntools:
**Stacked on #369** (base = `feat/use-case-convenience-plots`; it needs the `AAPredPlot.eval(kind="heatmap")` + `predict_group(band=True)` additions from that PR). Rebase onto `master` once #369 merges.

## What

Re-authors the flagship γ-secretase Use Case (`use_cases/use_case1_gamma_secretase.ipynb`) to drop its hand-rolled matplotlib "glue" and instead use the library convenience methods — covering the **full** prediction-plot surface in a few lines each:

| Figure | Before | After |
|---|---|---|
| Optimization heatmap (parts × scales) | `sns.heatmap(...)` + manual ticks/labels | `AAPredPlot().eval(kind="heatmap")` — boxes the best cell automatically |
| Prediction-eval heatmap (#features × #negatives) | `sns.heatmap(...)` block | `AAPredPlot().eval(kind="heatmap")` |
| Baseline comparison bars | manual `ax.bar` + per-bar `ax.text` + `axhline` | `AAPredPlot().eval(kind="comparison")` |
| dPULearn PCA | 3 fixed groups | 4 groups: `dPULearn.project(X_known)` → `dPULearnPlot().pca(df_pu_add=...)` |
| **new** score histogram | — | `predict_group(kind="hist", band=True)` — confidence-banded |
| **new** ranked candidates | — | `predict_group(kind="ranking", col_std=…)` — model-agreement error bars |

The per-protein SHAP feature maps already used `CPPPlot().feature_map(shap_plot=True)` and are unchanged; `SequenceFeature.feature_matrix(df_seq=…)` is available and used. The now-unused `import seaborn as sns` is removed.

## Why this shape

Per your steer — **master is the source of truth**, so `AAPredPlot` is used as it already exists on `master` (the VS Code plan to build it from scratch was based on a stale checkout). This notebook now exercises **all** the convenience capabilities from the appendix audit: 4 already shipped + the 2 added in #369.

## Verification

- Re-executed end to end via nbconvert (**~80s**, 0 errors) with fresh committed outputs.
- Visually confirmed each new figure: heatmap best-cell box, 4-group PCA (projected known negatives overlaid), banded histogram (blue→red across 20/50/80), ranking bars with error caps + cutoff lines.
- `nbformat.validate` passes.

Note: `use_cases/` is not in the nbmake CI gate; RTD renders committed outputs, which are refreshed here.

PR #381, opened 2026-07-09
docs(use-case): g-sec notebook — two appendixes (aap pipe + new aa convenience methods)

breimanntools:
Adopts the fuller γ-secretase use-case notebook (the version developed locally) and **replaces the appendix section with exactly two appendixes** that reproduce the five bespoke result figures from the same main-text data with far less user code:

- **Appendix 1 — the `aap` pipe**: discover → benchmark → explain, one call each.
- **Appendix 2 — the new `aa` convenience methods**: each bespoke figure is now a single call —
  `AAPredPlot.eval(kind='heatmap')`, `predict_group(kind='hist', band=True)`, `predict_group(kind='ranking')`, `dPULearnPlot.pca(df_pu_add=dpul.project(...))`, `CPPPlot.feature_map(shap_plot=True)` — vs the ~15-line manual matplotlib helpers.

Re-executed end to end against current master, so the plots carry all the recent fixes (auto_font grids, importance labels inside bars, interleaved SHAP bars). Both appendixes execute cleanly (3 + 7 figures, 0 errors); the sweep heatmap and 4-group PCA were visually verified to reproduce the main-text figures.

**Note:** this supersedes the compact single-appendix version from #370 (per the maintainer's decision to keep the fuller local notebook).


breimanntools (commented):
Will be checked locally. But first push your adjustments here

PR #382, opened 2026-07-09
feat(prediction): AAPred.eval(baseline=) compares CPP features vs composition baselines

breimanntools:
## What

Adds a `baseline` option to `AAPred.eval` that compares the bound (CPP) features against
simple, non-positional **composition baselines** built internally from `df_seq` — one call
now produces the whole "CPP vs baseline" comparison instead of a manual
`df_seq → X_baseline → model → score` loop.

```python
df_eval = aapred.eval(X, labels, df_seq=df_seq, baseline=["scale", "aac", "dpc"])
```

- `baseline=True` → the scale-composition baseline; a str / list selects among `scale`
  (`SequenceFeature.scale_composition`), `aac` (`aa_composition`), `dpc` (`dipeptide_composition`).
- Each baseline is cross-validated with the **same models and folds** and appended to `df_eval`
  under a new leading `features` column (`cpp` for the bound-feature rows).
- **Purely additive**: with `baseline=None` (default) `df_eval` is **byte-identical** to before
  (no `features` column). Baselines are cross-validation-only; the `holdout` principle stays
  `cpp`-only.

## Why

The composition featurizers (`scale`/`aac`/`dpc`/`kmer`) already ship on `SequenceFeature`, but
turning them into a baseline score was a manual, re-implemented loop. This puts the comparison
where prediction/evaluation lives, closing the measurable KPI of #335
(`AAPred.eval(baseline=True)` reproduces the use-case baseline row).

## Changes

- `_constants.py`: `COL_FEATURES`, `STR_FEATURES_CPP`, `LIST_BASELINE_KINDS`, `COLS_EVAL_PRED_FEATURES`.
- backend `aa_pred_eval.py`: `eval_models(dict_X_baseline=)` stacks CV-only baseline blocks
  (refactored the per-matrix scoring into `_eval_one`; no-baseline path is byte-identical).
- frontend `_aa_pred.py`: `eval(..., df_seq=, baseline=, list_parts=)` + `check_baseline` /
  `build_baseline_matrices` helpers + numpydoc.
- 12 unit tests incl. a **byte-identical parity** test (a baseline's rows equal a standalone
  `eval` of the same featurizer matrix).
- Example notebook `aapred_eval.ipynb` (re-executed) + release-notes entry.

## Verification

Local: 420 prediction + api_tests green · 12/12 new baseline tests · docstrings 0 defects ·
doc/signature drift 0 · notebook param-coverage (incl. prediction zero-gap) · method-spacing ·
backend-import-hygiene · flake8 clean. Full fast suite ran clean-under-`-x` (no failures); CI
runs the authoritative matrix.

## Scope

**Part of #335** (does not auto-close). The **ACC** (auto-cross-covariance) featurizer is
deferred to a follow-up: its `n_scales²` output needs a curated small default scale-set, a
separate design pass. #335 stays open to track it.


breimanntools (commented):
Looks good

PR #400, opened 2026-07-12
feat(prediction): AAPred.eval accepts a custom CV splitter with pooled scoring

### LINKED ISSUE #397: Support custom CV splitters (LeaveOneOut) + pooled scoring in AAPred.eval
## Problem

`AAPred.eval` hardcodes `StratifiedKFold(n_splits=n_cv)` in the backend
(`aaanalysis/prediction/_backend/aa_pred/aa_pred_eval.py::_score_cv`) and its
frontend guard `check_n_cv` caps `n_cv` at the smallest class count. On a
small, imbalanced benchmark — e.g. the γ-secretase substrate set (63 substrates
vs. 14 known non-substrates) — this makes **LeaveOneOut** cross-validation
impossible through the shipped API, so users drop back to hand-rolled sklearn:

```python
balanced_accuracy_score(y, cross_val_predict(SVC(kernel="linear"), X, y, cv=LeaveOneOut()))
```

There is a second, subtler gap: even if `eval` accepted a splitter, `_score_cv`
uses `cross_val_score` (score **per fold, then average**). Under LeaveOneOut
each test fold has a single sample, so per-fold `balanced_accuracy` is
degenerate and its average does **not** equal the study's metric, which pools
every held-out prediction and scores **once**. Closing this gap therefore means
adding both a custom-CV argument *and* a pooled-prediction scoring principle.

The absence forces every small-n evaluation in a use case to reimplement CV by
hand, which is exactly the boilerplate the prediction layer was meant to absorb.

## Goal

Let `AAPred.eval` accept an arbitrary scikit-learn CV splitter (including
`LeaveOneOut`) and score it via **pooled out-of-fold predictions**, reproducing
`balanced_accuracy_score(y, cross_val_predict(est, X, y, cv=cv))` exactly — with
the default integer-`n_cv` path (StratifiedKFold, per-fold scoring) left
byte-identical.

## Requirements

- [ ] `AAPred.eval` accepts a CV **splitter object** in addition to the int
      `n_cv` (e.g. a new `cv=` parameter, or `n_cv` accepting a splitter);
      backward compatible — int keeps the current StratifiedKFold behavior.
- [ ] Add a **pooled scoring principle** in
      `aa_pred_eval.py`: run `cross_val_predict` over the splitter, then apply
      each metric once on the pooled predictions. Tag these rows distinctly in
      the `principle` column (distinct from the existing per-fold `cv` principle)
      so both are legible in `df_eval`.
- [ ] Relax `check_n_cv` when a splitter is passed so it is not capped at the
      smallest class count; keep the cap for the int path.
- [ ] Frontend validation block for the new argument (splitter must expose
      `split`); backend trusts frontend.
- [ ] numpydoc update (named `Returns`, `Examples` include) documenting the new
      argument and the pooled `principle`.

## KPIs / Acceptance criteria

- [ ] `AAPred(models=[SVC(kernel="linear")]).eval(X, y, cv=LeaveOneOut(), metrics=["balanced_accuracy"])`
      returns a score equal within `1e-9` to
      `balanced_accuracy_score(y, cross_val_predict(SVC(kernel="linear"), X, y, cv=LeaveOneOut()))`.
- [ ] Default integer-`n_cv` `df_eval` output **byte-identical** to current
      (regression-tested).
- [ ] The γ-secretase Appendix-3 `balanced_acc` helper and its call sites are
      removed; the affected figures (optimization sweep, baseline-comparison
      bars) render **byte-identical**.
- [ ] Covered by ≥1 unit test and ≥1 property test; example notebook runs under
      the nbmake CI gate.

## Scope / non-goals

- Core feature (`topic:core`); no new dependencies, no `pro` gating.
- Not adding bootstrap confidence intervals or paired ΔMCC — that is #91.
- Not adding a train-only fold-augmentation set (the PU "fair fold") — that is
  intentionally left as user code for now.

## Dependencies

- relates #91 (repeated-CV / bootstrap CIs), #335 (baseline in `AAPred.eval`).
- surfaced while reducing the γ-secretase Appendix-3 boilerplate (cf. #305).

## Standards checklist

- [ ] **CONFIRM-FIRST**: touches the public `AAPred.eval` signature (additive,
      backward compatible) — flag for review.
- [ ] Frontend/backend split honored; validation block; backend trusts frontend.
- [ ] numpydoc docstring (named `Returns`, per-method `Examples` include).
- [ ] tests (unit + property/golden); reproducibility (`random_state` contract).
- [ ] no `print()` (`ut.print_out`); bare `ValueError`/`RuntimeError`; no
      `aaanalysis._utils.*` imports outside `utils.py`.

breimanntools:
## Summary

Closes #397.

`AAPred.eval` now accepts an arbitrary scikit-learn CV splitter via a new `cv=` parameter and scores it by **pooled out-of-fold prediction**, unblocking `LeaveOneOut` on small, imbalanced sets (e.g. the γ-secretase substrate set, 63 vs 14).

- **`cv=` splitter** (e.g. `LeaveOneOut()`, `StratifiedKFold(...)`). Chosen over overloading the integer `n_cv` so the scoring principle never depends on an argument's runtime type. `n_cv` keeps its exact meaning and is ignored when `cv` is given.
- **New `cv_pooled` principle** — `eval` runs `cross_val_predict` once over the splitter and applies each metric **once** on the pooled held-out predictions, reproducing `metric(labels, cross_val_predict(estimator, X, labels, cv=cv))`. `score_std` is `NaN` (a single estimate, like `holdout`). `roc_auc` uses the positive-class `predict_proba`; the other five metrics score hard labels.
- **Class-count cap relaxed for splitters** — `check_cv` validates a `split` method and bypasses the smallest-class-count cap that the integer `n_cv` path keeps. That cap is what made `LeaveOneOut` impossible through the shipped API.
- **Default (`cv=None`) output is byte-identical** — verified by md5 on the default+holdout and baseline paths; the per-metric row order is preserved. `baseline=` matrices are scored with the same splitter.

## Why pooled, not per-fold

The existing `cv` principle averages a per-fold score. Under `LeaveOneOut` each test fold holds one sample, so per-fold `balanced_accuracy` is degenerate (0/1) and its average does **not** equal the study metric, which pools every held-out prediction and scores once. Supporting `LeaveOneOut` correctly therefore required both the splitter argument *and* the pooled principle.

## Ripple

- Docstring (named `Returns`, documents `cv` + `cv_pooled`).
- Example notebook `examples/prediction/aapred_eval.ipynb` (new `cv=LeaveOneOut()` cell, re-executed with embedded outputs; prediction notebooks are held to zero undemonstrated params).
- Release notes (Unreleased).
- `docs/module_map.md` — added the `prediction` subpackage (AAPred/AAPredPlot/ReliabilityModel), which was missing from the dataflow map.

## Tests

- `TestAAPredEvalCV` — principle tagging, `NaN` std, LeaveOneOut bypasses the cap, StratifiedKFold splitter (hypothesis), roc_auc via proba, `cv=None` byte-identical, reproducibility, invalid-splitter negatives.
- `TestAAPredEvalCVGoldenValues` — pooled `LeaveOneOut` `balanced_accuracy` equals `balanced_accuracy_score(y, cross_val_predict(SVC(kernel="linear"), X, y, cv=LeaveOneOut()))` within `1e-9`; pooled scores equal the manual `cross_val_predict` reference for all label metrics.

## Scope note

The KPI item to remove the γ-secretase Appendix-3 `balanced_acc` helper touches `use_case1_gamma_secretase.ipynb`, which has an in-flight PR (#389) on exactly that file. This PR ships the API that unblocks that cleanup; the notebook edit belongs in #389 or a follow-up rather than colliding here.


breimanntools:
Skipping manual review — all **required** checks are green and the automated gates (/review, /security-review, docstring + ADR checkers, 6296-test local run) passed.

Note on the failing `visual-regression` check: it is **advisory (not a required status check)** and is **not caused by this PR** — the diff touches zero plotting/baseline files. The two failures (`test_feature_map_baseline`, `test_heatmap_baseline`) are pre-existing feature_map/heatmap dimension mismatches inherited from the cpp cell-size change (#387) on master (an environment-sensitive render: 'Arial not found → sans-serif fallback'). The baselines should be refreshed in a separate follow-up on master, independent of this feature.

PR #404, opened 2026-07-13
fix(cpp-plot): keep feature_map/heatmap furniture below the grid, survive tight_layout

breimanntools:
## Summary

Fixes user-reported layout bugs in the composite CPP plots (`CPPPlot.feature_map` / `heatmap`).

- **Furniture no longer lands on the heatmap.** These methods hand-place their colorbar and legends below the grid in figure coordinates. The standard notebook idiom `plt.tight_layout(); plt.show()` re-packed every axes and dropped that furniture back onto the heatmap. Both methods now manage their own layout and **neutralize `tight_layout` on the returned figure** (`_freeze_composed_layout`, like seaborn's clustermap). `savefig(bbox_inches="tight")` and a plain `plt.show()` are unaffected. `set_in_layout(False)` was tried first but does **not** stop `tight_layout` in matplotlib 3.11.
- **No-sequence TMD/JMD bar** is drawn as a solid, sequence-band-like track (height factor 2.5) instead of a razor-thin strip; part labels sit a matching gap below it.
- **Explicit `figsize` still keeps its exact size** (`grow_to_fit`/`align_bottom_furniture_` stay gated on the sizer path); `figsize=None` remains the auto-sizing default.
- **ShapModel `add_feat_impact`** "grouping might be invalid" advisory is now a verbose-gated `ut.print_out` instead of a `warnings.warn` traceback — clean notebook output.

## Ripple
- Docstrings: `feature_map`/`heatmap` Notes document the self-managed layout.
- Release notes: `Fixed` entry.
- The three example notebooks re-executed with clean output (`sample_positions` → `samples`; no fallback/grouping warnings).
- +2 test classes: `TestTightLayoutFreeze`, `TestNoSequenceBarHeight`.

## Verification
- Rendered default / no-top / narrow & large manual-`figsize` / post-`tight_layout` cases for both methods — furniture below the grid in every one.
- 471 plot tests green (465 existing + 6 new); docstrings 0 defects / 0 drift; notebook + param-coverage checkers clean; no new pyright diagnostics.


breimanntools (commented):
push to master and refresh notebooks (esp. feature_map)

PR #417, opened 2026-07-14
feat(data): bake a gene column into every dataset (drop add_metadata)

### LINKED ISSUE #410: Bundled-dataset display metadata (`gene` / `display_name` / `organism`) via `load_dataset(add_metadata=True)`
## Problem
`load_dataset("DOM_GSEC")` returns `entry`/`sequence`/`label`/domain coords but **no human-readable
metadata**, which is why the g-sec notebook carries a 130-entry hardcoded `DICT_GENE` gene map. Any plot
or table that wants readable labels (ranking, clustermap, SHAP maps) reconstructs it by hand, and the
just-shipped `sample`-by-name resolver (keys on a df_seq `name` column) has nothing to resolve against for
the bundled datasets.

## Goal
Make the bundled datasets self-describing — opt-in metadata columns keyed on a canonical identifier — without changing existing default return shapes.

## Requirements
- [ ] `load_dataset(..., add_metadata=False)` (default off → existing shape unchanged); when `True`, add `gene`, `display_name`, `organism` (canonical `entry` stays the identifier).
- [ ] Coordinate the resolver column: the shipped `sample`-by-name resolver keys on `name` — either populate `name`, or extend the resolver to consult `gene`/`display_name` (decide in the PR). Resolution order: exact `entry` → unique `gene` → unique `display_name` → else ambiguity `ValueError`.
- [ ] Ship the metadata for at least `DOM_GSEC` / `DOM_GSEC_PU` (source: UniProt gene names).
- [ ] Touches `aaanalysis/_data/` — **CONFIRM-FIRST**; call out the exact files.

## KPIs / Acceptance criteria
- [ ] `load_dataset("DOM_GSEC")` default output **byte-identical** to current (regression-tested); `add_metadata=True` adds exactly the documented columns.
- [ ] `sample_kws={"sample":"APP"}` resolves on `DOM_GSEC` end-to-end (test).
- [ ] Ambiguous/absent gene raises a clear `ValueError` naming the offending token.

## Scope / non-goals
- Only bundled datasets; not a UniProt fetch path (that's downstream / `pro`).

## Dependencies
- relates the shipped `sample`-name resolver (branch `feat/sample-name-resolve`); **#336**.

## Standards checklist
- [ ] **CONFIRM-FIRST: `_data/`** · numpydoc · regression test on default shape · `ValueError` (not custom).

---

breimanntools:
Closes #410

## What (reworked per review)
Replaced the opt-in `add_metadata` parameter with a first-class **`gene` column baked into every bundled dataset table**, right after `entry` — cleaner and self-describing, no flag, no separate metadata file.

- **Domain datasets** (`DOM_GSEC` / `DOM_GSEC_PU`): `gene` is the UniProt **primary gene symbol**, 100% coverage (126/126 and 694/694), fetched by a committed one-off build script `.github/scripts/add_gene_names_to_datasets.py`. Matches the study's curated gene map exactly (0 differences).
- **Amino-acid / sequence datasets** (`AA_*` / `SEQ_*`): entries are synthetic ids with no gene → `gene` is a positional `name_<row>` placeholder (per the agreed peptide-level fallback).

The bundled `gene` column makes gene-based `sample` selection work out of the box: `resolve_sample_entry` (`get_seq_kws` + `sample_kws`) consults optional `gene` / `display_name` / `name` columns of `df_seq` (order `entry` → `gene` → `display_name` → `name`), so `sample="APP"` resolves on `DOM_GSEC`.

## Byte-identity + safety
- **Every other column and value is byte-identical to master** (verified per-column across all 14 tables — only the `gene` column is inserted after `entry`).
- The **CPP regression anchor** (DOM_GSEC) and the **integration suite** (42 tests) pass unchanged — `gene` is ignored by `get_df_parts` / the CPP spine.
- AA **windowing** rebuilds per-position entries and does not carry `gene` (documented).

## CONFIRM-FIRST `_data/`
This touches all bundled dataset TSVs under `aaanalysis/_data/benchmarks/` (the `gene` column) and **removes** the earlier `metadata/DOM_GSEC.tsv`. Packaged by the existing setuptools `_data/**/*.tsv` rule (no `pyproject` change).

## Note on history
Two commits (the original `add_metadata` approach, then this rework); the **net PR diff is the clean baked-`gene`-column feature** — `add_metadata` is fully removed (0 occurrences). Say the word if you'd like it collapsed to a single commit (force-push).

## Verified locally
`data_handling` + `sequence_feature` + `cpp_plot_sample` + param-coverage / notebook-param-coverage (1002 tests) + nbmake + CPP regression anchor + docstring checker — all green.


breimanntools (commented):
Do you have all names? Pleaese simply make a script to fetch for each existing dataset in load_scales the respective gene name for the entry and add gene (name) to all tables. This is much easier and cleaner. Then we can remove this add_meta_data. Add this after entry to the table and make sure that all other infos are byte identical. If name is not possible (use name_XX X is number of respective peptdie) for peptide level data.

PR #452, opened 2026-07-16
docs(thumbs): render the P2 logo wide, per the g-sec use case settings

breimanntools:
Reworked after review: **the earlier collision-based aligner is reverted** — no plot-code change remains. The net diff is `thumb2.py` + the regenerated card.

**The real problem was the thumbnail's aspect, not the plot code.** A sequence logo reads wide and short. thumb2 asked for a `7x7` square, which overrode `single_logo`'s own `(8, 3.5)` default: that stretched the letters vertically and squeezed 25 position ticks into a width where the JMD-N/JMD-C labels had nowhere to sit but on top of them.

**Fix.** Take the gamma-secretase use case settings: `plot_settings(font_scale=0.85)` and `figsize=(9, 3)` — its `figsize_per_logo`, that figure just stacks three logos. Dropping the `fontsize_tmd_jmd=15` / `weight_tmd_jmd="bold"` override leaves the part labels **directly under the bar** at the default, slightly smaller size.

At a sensible aspect the labels clear the ticks with room to spare, so nothing in the library needs to move.


breimanntools (commented):
NO!!!! This is wrong. Keep TMD, JMD drietly under the bar but a bit smaller. Also adjsut the size of the figure wider than high to have a better logo. Simply take the settings like in the g-sec use case notebook (but then 1/3 as we have three plots there) or as in the aalogoplot example notebook!!!

PR #453, opened 2026-07-16
docs(protocol1): teach the CPP concepts with six feature maps

breimanntools:
P1 drew a **single** feature map, and described compositional vs positional signatures only in prose, so the concepts that actually decide a CPP run were never shown. Each concept now gets its own map, with everything else held fixed so a map isolates exactly one decision.

| Map | `split_kws` / filter | What it teaches |
| --- | --- | --- |
| 1. raw signature | default (all three split types) | what `run` alone leaves you: 50 features, 30 subcategories |
| 2. refined signature | same, after `CPP.simplify` | the second CPP step: 50 → 35 features, 19 subcategories |
| 3. compositional | `Segment`, `n_split_max=1` | part-wide bands: *which* property, position-agnostic |
| 4. positional | `Segment`, `n_split_min=2, n_split_max=6` | narrow cells: *where* in the part the signal sits |
| 5. pattern | `Pattern` | offsets from a terminus, not contiguous chunks |
| 6. one family | filter `category` | whether a block is coherent within one AAontology family |

**The payload is maps 3 and 4.** Compositional renders as solid part-wide bands (Charge across *all* of `jmd_n`, α-helix across *all* of the `tmd`) because that is exactly what one `Segment(1,1)` average covers. Positional breaks into narrow cells concentrated at the TMD/JMD-C boundary. Same data, same scales: only the split rule changed. That is the difference between *which property* and *where*, made visible instead of asserted.

Terminology follows the `CONTEXT.md` glossary entry *compositional vs positional (CPP strategy)*.

**Notable change.** `run()`'s output previously got overwritten by the `simplify` step, which made the before/after contrast impossible to draw. It now survives as `df_feat_run`, with the ranked copy under its own name — `simplify` still consumes an unranked frame, which `add_feat_importance` requires (it refuses a frame that already carries the column).

MAP 2's cell drops a now-duplicate import and gains the `tight_layout` the house style wants.

**Verified.** Re-executed end to end with embedded outputs: **6 maps, 0 errors**; `nbformat.validate` OK; tables all via `display_df`, no bare `df.head()` / `print(df)`.


breimanntools (commented):
push so that I can see!

PR #464, opened 2026-07-18
ci(types): pyright ratchet gate (#442) + docs: four evaluation regimes (#461)

### LINKED ISSUE #442: Make the pyright ratchet gate (block new diagnostics) + pin pyright + drop the toothless mypy step
## Problem

`.github/scripts/check_pyright_budget.py` already implements a no-regression ratchet (exit 1 when diagnostics exceed the committed `887` baseline), but `pyright.yml` wraps it in `continue-on-error: true`, so it **never gates** — new type errors accumulate silently and the 887 baseline becomes fiction. Separately, `codeql_analysis.yml` runs `mypy aaanalysis/__init__.py --follow-imports=skip` (one file, imports skipped) which is effectively meaningless *and* contradicts `sharp-edges.md` ("mypy is not used") and the project's choice of pyright as the checker. Agent-generated code is the main source of plausible-but-wrong type drift, so an un-gated checker misses exactly what it is for.

## Goal

Turn the existing pyright ratchet into a real **no-regression gate** — block *increases* above the baseline, never require clean — reproducibly, without freezing velocity.

## Requirements

- [ ] Remove `continue-on-error: true` from the **budget-check** step in `pyright.yml` (keep it on the raw `pyright` step, which may still print). **CONFIRM-FIRST (`.github/workflows/`).**
- [ ] Pin the pyright version in the install/run so the baseline count is reproducible.
- [ ] Remove the `mypy aaanalysis/__init__.py --follow-imports=skip` step from `codeql_analysis.yml`. **CONFIRM-FIRST.**
- [ ] Update `sharp-edges.md` so it matches CI (mypy removed; pyright ratchet gates no-new-diagnostics).
- [ ] Document the decision in the repo's design docs: the pyright checker moves from advisory-only to a no-regression ratchet (still not a clean/strict gate). Keep the narrow `# pyright: ignore[rule]` (+ inline reason) escape valve.
- [ ] Maintainer: add the pyright job to branch-protection required checks.

## KPIs / Acceptance criteria

- [ ] A PR that pushes pyright diagnostics above 887 fails CI; a PR at/below 887 passes.
- [ ] Bumping pyright is a deliberate, re-baseline change (version pinned).
- [ ] No `mypy` invocation remains in any workflow; `sharp-edges.md` matches CI.

## Scope / non-goals

- **Not** a clean/strict gate (0 errors) — velocity stays; merging at 887 is allowed.
- No `pyrightconfig.json` scope expansion here (public + non-backend stays).

## Standards checklist

- [x] CONFIRM-FIRST: `.github/workflows/`
- [x] records the advisory-to-ratchet decision in the design docs

Part of the agent-readiness & release-maturity hardening epic.

### LINKED ISSUE #461: Document the four evaluation regimes (exploratory / nested-CV / final fit / external test)
## Problem

`aaanalysis.pipe.find_features(selection_scope="global")` — the default —
selects CPP features on the full labeled set and then cross-validates the model
on that fixed feature matrix, so every reported score (the whole `df_eval`
sweep) is post-selection / in-sample optimistic. The code warns about this, but
on the small biological datasets AAanalysis targets, users read the number as a
held-out generalization estimate and over-trust it. There is no user-facing
documentation that separates the distinct things a CPP / model score can mean.

## Goal

A user-facing docs section that clearly distinguishes the four evaluation
regimes, so a reader knows which one produced any given number and never
mistakes an exploratory score for a generalization estimate.

## Requirements

- [ ] New docs section covering the four regimes:
  - **exploratory feature discovery** — `selection_scope="global"`; a relative
    ranking signal, not a generalization estimate (the determinant-discovery use).
  - **honest (nested-CV) evaluation** — `selection_scope="fold"`; selection
    re-runs inside each fold, a leakage-free held-out estimate.
  - **final model fitting** — refit on all data for the deployable model; its
    training score is an artifact of fitting, not an evaluation number.
  - **external test-set evaluation** — the final model scored on data held out
    from all selection and CV.
- [ ] Cross-linked note in the `find_features` example notebook demonstrating
  `global` vs `fold` side by side, stating which regime each `df_eval` score
  belongs to.
- [ ] Canonical definitions anchored in the project glossary (already added:
  `selection_scope`, `evaluation regime`).

## KPIs / Acceptance criteria

- [ ] Section renders on RTD and is reachable from the feature-selection /
  find_features docs.
- [ ] The `find_features` example shows both regimes with the
  score-interpretation caveat inline.

## Scope / non-goals

- Documentation only; the `selection_scope` mechanism already ships.
- Not the nested-CV Monte-Carlo engine (#276).

Part of the v1.1 release-hardening (#449). Related: #411 (selection_scope,
implemented).

breimanntools:
Two release-hardening changes (built together; say the word if you'd rather split into separate PRs).

**#442 — pyright ratchet gate.** Turns the existing pyright budget check into a real no-regression gate (fails when diagnostics rise above the committed `.github/pyright_baseline.txt`, allows merging at/below), pins `pyright==1.1.410` so the baseline is reproducible, and removes the toothless mypy step. Gate logic verified (887 → pass, 888 → fail). Rules/docs synced.
_Maintainer follow-up:_ add the `pyright` job to branch-protection required checks — the check was renamed `Type Check (advisory)` → `Type Check (ratchet)`, so update any rule referencing the old name.

**#461 — evaluation-regimes docs.** New user-facing chapter separating the four regimes (exploratory `global` / honest nested-CV `fold` / final refit / external test), wired into the feature-selection concept page and cross-linked from the `find_features` example (re-executed). Docs build clean.

Closes #442
Closes #461


breimanntools (commented):
okay

PR #500, opened 2026-07-18
test: deterministic parallel-execution conformance (n_jobs invariance)

### LINKED ISSUE #486: Add deterministic parallel-execution conformance tests (n_jobs invariance)
## Problem

The package promises deterministic, seed-stable scientific outputs, and `n_jobs`
parallelism is used across CPP. But there is no test that the *same* computation gives
the *same* result at `n_jobs=1` and `n_jobs>1`, or chunked vs unchunked — so a
parallelism change (or a non-deterministic third-party op) could silently alter a
scientific output and no gate would catch it. Issue #339 (an `n_jobs>1` crash) shows the
parallel path is under-guarded.

## Goal

Add a conformance test suite asserting that parallel and chunked execution produce
outputs equivalent to the single-worker, in-memory path within a documented numerical
tolerance, with seeds propagating correctly.

## Requirements

- [ ] Parametrised tests comparing `n_jobs=1` vs `n_jobs>1` (and chunked vs unchunked,
      once chunking lands) on the CPP feature path and the affected predictors.
- [ ] Documented numerical tolerance per compared output (exact where integer/ordering,
      `atol/rtol` where floating-point reductions differ by association).
- [ ] Assert seed propagation: a seeded parallel run reproduces across repeats and
      matches the serial run within tolerance.
- [ ] Identify and document any third-party op that is legitimately non-deterministic
      (so it is a known, pinned exception rather than a silent one).
- [ ] Wire into the test suite so a parallelism change that alters output fails CI.

## KPIs / Acceptance criteria

- [ ] `n_jobs=1` and `n_jobs>1` produce `df_feat` equal within the documented tolerance
      on a seeded fixture (asserted); output ordering is stable across worker counts.
- [ ] A seeded parallel run is reproducible across repeated invocations (asserted).
- [ ] The suite fails if a deliberately-introduced order-dependent change is made
      (guard is real, verified by a mutation in review).
- [ ] Runs in the existing test matrix; no new runtime dependency.

## Scope / non-goals

- **Core** — tests only; may add small tolerance documentation.
- Reproducibility across workers is a scientific-quality responsibility that is
  AAanalysis's own (not a border question). Distributed/cluster execution stays in
  ProtXplain.

## Dependencies

- Relates the `n_jobs>1` crash (#339), the `n_jobs` contract, and the chunked-features
  issue.

## Standards checklist

- [ ] Frontend/backend split honored; n/a (tests)
- [ ] CONFIRM-FIRST? — no
- [ ] numpydoc n/a; tolerance documented
- [ ] tests (conformance: parallel/chunked equivalence, seed propagation); reproducibility
- [ ] no `print()`; bare `ValueError`/`RuntimeError`; no `aaanalysis._utils.*` outside `utils.py`

breimanntools:
Adds `tests/integration/test_parallel_conformance.py`: a metamorphic guard that the same computation returns the same scientific output regardless of how it is parallelized (`n_jobs`) or chunked (`n_batches` / `n_sample_batches`), so a parallelism change (or a non-deterministic third-party op) that silently altered output fails CI.

## Covered (byte-identical, `n_jobs=1` vs `n_jobs>1`)
- `CPP.run` `df_feat` (values + row order), reproducible across repeats and vs serial
- `SequenceFeature.feature_matrix` `X`
- `comp_auc_adjusted` AUC vector
- `dPULearn.eval` `df_eval` (the KLD parallel path, `comp_kld=True`)
- `CPPGrid` per-config `df_feat`

## Documented tolerance per output
- `n_jobs` (worker count) and `n_sample_batches` (sample-axis chunking): **exact** (byte-identical).
- `n_batches` (scale-axis chunking): **exact on the selection-driving columns** (feature set, row order, `abs_auc`, `abs_mean_dif`) but perturbs **only** `p_val_fdr_bh`, because Benjamini-Hochberg FDR correction is by definition computed across the whole candidate set. This is a known, expected batch-dependence (documented on `CPP.run`), not nondeterminism — feature selection and ranking are unaffected.

No third-party op in the covered paths is legitimately non-deterministic: joblib splits an independent per-feature loop and the results are concatenated in a fixed chunk order, so determinism is structural. The guard is exact-equality based, so any change that reorders or reduces across the worker split breaks it immediately (verified in review by reversing the parallel concat order → the conformance tests fail).

Tests-only change; no new runtime dependency; runs in the existing Integration & E2E matrix.

Closes #486

### DIFF
--- a/tests/integration/test_parallel_conformance.py
+++ b/tests/integration/test_parallel_conformance.py
@@ -1,0 +1,211 @@
+"""This is a script to test deterministic parallel-execution conformance (n_jobs invariance).
+
+Integration tier: a *metamorphic* guard that the same computation returns the same
+result regardless of how it is parallelized (``n_jobs``) or chunked (``n_batches`` /
+``n_sample_batches``). ``n_jobs`` only splits the per-feature work loop across
+workers and concatenates the per-chunk results in a fixed chunk order, so no
+cross-worker reduction ever changes a value: the serial and parallel paths are
+byte-identical. This suite pins that contract so a future parallelism change (or a
+non-deterministic third-party op) that silently altered a scientific output fails
+CI. It complements, not duplicates, the unit ``test_n_jobs`` contract tests (which
+check the ``check_n_jobs`` / ``resolve_n_jobs`` normalizer, not output equivalence).
+
+Documented tolerance per compared output (numerical-equivalence policy tiers):
+
+* ``n_jobs`` (worker count) invariance -> EXACT (T1), byte-identical:
+    - ``CPP.run`` ``df_feat`` (values AND row order)
+    - ``SequenceFeature.feature_matrix`` ``X``
+    - ``comp_auc_adjusted`` AUC vector
+    - ``dPULearn.eval`` ``df_eval`` (the KLD parallel path, ``comp_kld=True``)
+    - ``CPPGrid`` per-config ``df_feat``
+* ``n_sample_batches`` (sample-axis chunking) invariance -> EXACT (T1): ``CPP.run``
+  ``df_feat`` is byte-identical to the unchunked run.
+* ``n_batches`` (scale-axis chunking) invariance -> EXACT on the *selection-driving*
+  outputs (selected feature set, row order, ``abs_auc``, ``abs_mean_dif``) but NOT on
+  ``p_val_fdr_bh``: the Benjamini-Hochberg FDR correction is by definition computed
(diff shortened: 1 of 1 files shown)


breimanntools:
Skipping manual review — automated gates green (`/review` approve, no changes required; `/security-review` clean, tests-only diff). Order-mutation check verified the guard bites (reversing the parallel concat order fails the AUC conformance tests). Arming auto-merge (merge commit); it lands once CI is green.

Issue #509, opened 2026-07-22
pipe: parity anchor for predict_samples, the ≤10-line KPI, and param-coverage gating for ap notebooks


breimanntools:
## Problem

The `aaanalysis.pipe` (`ap`) layer and `SequenceFeatureTransformer` shipped in v1.1.0 (see #241), but three guarantees from the original golden-pipelines spec are unenforced:

- `ap.predict_samples` is the only End without a byte-identical parity test, so nothing pins its defaults to the explicit `feature_matrix` → estimator → `cross_validate` path.
- The "one call reaches a CV score in ≤10 lines" ergonomics promise is asserted nowhere.
- The notebook parameter-coverage gate iterates `aaanalysis.__all__`, which by design excludes `pipe` — so the five `examples/pipe/*.ipynb` are the only example notebooks in the repo with no coverage gate, and two params (`obtain_samples.pos_col`, `plot_eval.score_col`) are already undemonstrated.

## Goal

Close the three enforcement gaps in the shipped `ap` layer without changing any behaviour.

## Requirements

- [ ] Byte-identical parity test for `ap.predict_samples` against the explicit `SequenceFeature.feature_matrix` + estimator + `cross_validate` chain at default settings (fixed `random_state`).
- [ ] Decide and record whether parity is definable for `find_features` at `search="balanced"`/`"exhaustive"`: add a pinned-config anchor, or state in the pipeline docstring that parity holds only at `search="fast"`.
- [ ] End-to-end ergonomics test asserting the `load_dataset` → `find_features` → `predict_samples` path is ≤10 statements, so the KPI is regression-guarded.
- [ ] Extend `tests/unit/api_tests/test_notebook_param_coverage.py` to also enumerate `aaanalysis.pipe.__all__` (keeping `pipe` out of the top-level `__all__`), with zero permitted gaps.
- [ ] Demonstrate `obtain_samples(pos_col=...)` and `plot_eval(score_col=...)` in their notebooks; pass `predict_samples`' `list_df_feat` by keyword at least once.
- [ ] Add a narrative "Golden pipelines" page under `docs/source/index/usage_principles/`.

## Non-goals

No behaviour change, no new pipeline, no new dependency.

PR #533, opened 2026-09-15
test(pipe): enforce the golden-pipeline parity, ergonomics and notebook gates

breimanntools:
Addresses #509.

Closes the enforcement gaps in the shipped pipeline layer. The parity anchor compares the full
fitted state of each estimator, including the arrays inside each tree, with guard tests proving
it detects a different fit. The ten-statement ergonomics promise is asserted against the snippet
parsed from the documentation page, so the two cannot drift. The notebook parameter gate now
covers the pipeline namespace, skipping entries that are stubs without the optional extra.

Also adds a narrative Golden Pipelines page, and corrects a false claim in it: not every
pipeline returns a three-slot result, since the plotting helper returns a list of figures.

No behaviour changed; the only library edit is a docstring that scopes the parity claim to the
fast search setting.

Not included, by decision: deciding whether parity is definable for the other search settings,
so the issue stays open.

https://claude.ai/code/session_015j7gVR2ba1zaYMmKxixdwF


breimanntools (commented):
Fine

Issue #455, opened 2026-07-17
Bring protocols 2-10 up to the Protocol 1 teaching standard (rubric + execution gate)


breimanntools:
## Problem

The ten workflow protocols (`protocols/protocol1`–`protocol10`) already share one
narrative skeleton — every one has *When to use it → Input → Run → How to interpret →
Common mistakes → Next step* (verified across all ten). The gap is not structure, it is
**teaching depth**.

Only **Protocol 1 (CPP signature)** currently *shows* each concept it names: it renders the
same signature six ways — raw `run` vs after `simplify`, compositional vs positional,
`Pattern` split type, and a category-filtered view — so every decision the method makes is
visible, not just asserted. The others mostly *tell*: P2, P3, P4, P5, P7, P8 and P10 each
carry a **single** visualization, so the reader is asked to take the prose on faith. P4
"Prediction levels", for instance, describes the residue / domain / protein levels in text
but visualizes only once.

Two costs now:

- Protocols are the RTD pages a new user reaches first. An under-illustrated protocol is the
  difference between "I can see what CPP is doing" and "I'll try something else".
- Protocol notebooks have **no execution gate**. `pytest --nbmake` covers only `tutorials/`
  and `examples/`; `integration_e2e.yml` runs `tests/integration tests/e2e` (Python tests),
  not the `.ipynb`. RTD renders the *committed* outputs without executing them, so a protocol
  can silently rot — stale outputs, a broken cell — and still ship.

## Goal

Bring protocols 2–10 up to the teaching standard set by Protocol 1 (each core concept shown
with a dedicated, contrasting visualization, and the demonstrated method's full public
surface exercised), and add an execution gate so protocols stay runnable — without changing
the existing narrative skeleton.

### Define the standard (do first)
- [ ] Write a short **protocol quality rubric** under `docs/guides/`, distilled from
      Protocol 1: (a) each named concept gets its own visualization that contrasts it against
      the alternative; (b) every public parameter of the demonstrated method is passed by
      name, grouped; (c) the existing section skeleton is kept; (d) every DataFrame via
      `aa.display_df(df, n_rows=10, show_shape=True)`, every plot cell ends
      `plt.tight_layout(); plt.show()`; (e) executed outputs committed.

### Keep them runnable
- [ ] Add `protocols/` to the local nbmake gate (the `workflow.md` recipe) and to the
      notebook-execution CI step alongside `tutorials/` / `examples/`.
[...]

PR #534, opened 2026-09-15
docs(protocols): bring protocols 2 and 3 to the teaching rubric

breimanntools:
Addresses #455.

Brings protocols 2 and 3 up to the teaching rubric. Each concept that was previously asserted in
prose now has a figure: an information-content logo beside the probability logo, a shuffled
baseline, a like-for-like group comparison, the distance band, the source-strategy composition,
motif-matched sampling, and the similarity filters. Common mistakes are demonstrated rather than
described.

Every demonstrated method passes its public parameters by name across its calls, tables use the
house presentation, and every plot cell ends correctly. Both notebooks were re-executed: no error
output, no stderr, kernel name unchanged.

A review round corrected factual claims, including the geometry of an even-numbered window and
the anchor positions of two substrates, which changed derived tables and figures.

Not included, by decision: protocols 4 to 10 and the execution gate in CI, which needs a workflow
edit, so the issue stays open.

https://claude.ai/code/session_015j7gVR2ba1zaYMmKxixdwF


breimanntools (commented):
This is very shallow. But if the implremenation is fine and nothing else is behind it, then merge it

